PR #8, opened 2025-06-30
Edit landau.plot.plot_1d_T_phase_diagram to return the axes and transition temperatures

prabhath-c:
Added functionality to `plot_1d_T_phase_diagram()` to be able to return the axes and list of transition temperatures.
Added more optional arguments like `mark_transitions=True`, to choose whether to mark the transition temperatures on the plot.


landau/plot.py
```diff
@@ -362,26 +362,70 @@ def plot_1d_mu_phase_diagram(df):
...
+        If True, all transition temperatures are marked on the plot.
+    return_transitions : bool, default False
+        If True, returns a list of transition temperatures alongside the Axes.
+
+    Returns
+    -------
+    ax : matplotlib.axes.Axes
+        The Axes object with the phase diagram plot.
+    transitions : list of float, optional
+        List of transition temperature values (only returned if return_transitions is True).
```
pmrv, line 394:
```suggestion
```

I would prefer not to return the temperatures here again, to keep the API focused.  Callers of the plot function can use small snippet like

```python
df.query('border')['T'].unique()
```

to obtain the same result anyway.  If you'd like that as a short cut in the library, I think a better place would be in `calculate.py`.

.gitignore
```diff
@@ -0,0 +1,2 @@
+__pycache__/
+*.pyc
```
pmrv, line 2:
```suggestion
```

I prefer to keep `.gitignore` outside of git, because everyone tends to add their own IDE files and such, e.g. I use this one for landau

```
.gitignore
build
dist
landau/_version.py
*checkpoints*
*__pycache__*
*.egg-info
```

landau/plot.py
```diff
@@ -362,26 +362,70 @@ def plot_1d_mu_phase_diagram(df):
...
+        show=True, 
+        mark_transitions=True,
+        return_Tt=False):
+    """
+    Plot a one dimensional equipotential phase diagram as a function of temperature.
+
+    Parameters
+    ----------
+    df : pandas.DataFrame
+        Input data containing columns for temperature ('T'), potential ('phi'),
```
pmrv, line 377:
```suggestion
    Args:
        df (pandas.DataFrame):
            Input data containing columns for temperature ('T'), potential ('phi'),
```

We use [google style](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html#example-google) in pyiron usually.  (It's less typing too :p)

pmrv (changes requested):
Thanks for your contribution!  I left some smaller suggestions, I'd like you to fix, but in general looks good to me, thanks!

prabhath-c:
Added functionality to return axes for `plot.plot_1d_mu_phase_diagram`. Additionally, attempted to fix the extra line being plot between two `border` points.

I tried to split the dataframe into sections based on the border points, and plot each separately.
It works fine, except for one small issue- there is now a small gap between two sections because of a small issue when sectioning.

pmrv:
Closed in favour of #12 #11

PR #16, opened 2025-08-12
Add new poly_method based on TSP

pmrv:
Solve the Traveling Salesman Problem on the boundary points of phases. It is slower than the segments or even concave methods, but much more stable.


pmrv:
Missing input alarms and then can be merged.

PR #35, opened 2025-10-22
Add softplus interpolation over concentration

prabhath-c:
Added functionality for using the Softplus function for interpolation. Especially for line phases, it shows a much better fit. Below are the results of the `check_concentration_interpolation` function for the Al-Mg phases at 1000K:

grafik

### DIFF
--- a/landau/interpolate.py
+++ b/landau/interpolate.py
@@ -244,7 +244,7 @@ class SoftplusFit(ConcentrationInterpolator, TemperatureInterpolator):
     """Number of softplus terms to fit."""
     loss: Literal["linear", "soft_l1", "huber", "cauchy", "arctan"] = "soft_l1"
     """Loss function for robust fitting."""
-    max_nfev: int = 100000
+    max_nfev: int = 100
     """Maximum number of function evaluations."""
 
     def fit(self, x, y) -> Callable[[float], float]:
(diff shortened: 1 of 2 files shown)


landau/interpolate.py
```diff
@@ -228,3 +229,78 @@ def interpolation(t):
             return f
 
         return interpolation
+
+@dataclass(frozen=True, eq=True)
+class SoftplusFit(ConcentrationInterpolator, TemperatureInterpolator):
+    n_softplus: int = 2
+    """Number of softplus terms to fit."""
+    loss: str = "soft_l1"
```
pmrv: Document possible values using `typing.Literal` or don't expose.

landau/phases.py
pmrv: Just call and return the new interpolator field here as I wrote above.

landau/interpolate.py
```diff
@@ -123,6 +123,7 @@ class RedlichKister(ConcentrationInterpolator):
 
     def __post_init__(self):
         assert self.nparam > 0, "Must fit at least one parameter!"
+        # FIXME: Was earlier in phases.py- object.__setattr__(self, "nparam", max(1, self.nparam - 2))
```
prabhath-c: This was a default value handed in `phases.py`. Might need to handle it in here now.

pmrv (changes requested):

PR #41, opened 2025-11-27
Add new interpolator

pmrv:
The purpose of this interpolator is to provide a guaranteed convex output function that is stable against noise and more amendable to certain types of free energies.  This also refactored the handling of the concentration interpolation in the `SlowInterpolatingPhase`, so that an interpolator under user control may be passed in rather than being chosen internally.


pmrv:
@prabhath-c You can open PRs directly again main next time, there's no issue or danger with this.

PR #82, opened 2026-04-28
Add analytical Jacobian to softplus interpolation fitting

pmrv:
## Summary
Improved the numerical stability and efficiency of the softplus interpolation fitting by adding an analytical Jacobian computation and implementing a numerically stable sigmoid function.

## Key Changes
- **Added numerically stable sigmoid function**: Implements separate computation paths for positive and negative values to avoid overflow in exponential calculations
- **Implemented analytical Jacobian**: Computes exact derivatives of residuals with respect to all parameters (amplitude, slope, and offset for each softplus term, plus global offset), enabling more efficient and accurate optimization
- **Code refactoring**: 
  - Cached `self.n_softplus` as local variable `n` for cleaner code
  - Changed accumulation from `out += ...` to `out = out + ...` for consistency
  - Updated `least_squares` call to pass the `jac` parameter
  - Removed redundant `loss` and `max_nfev` parameter positioning

## Implementation Details
The analytical Jacobian leverages the chain rule for each softplus term:
- For `a * softplus(b * (xn + c))` with `t = b*(xn+c)`:
  - ∂/∂a = softplus(t)
  - ∂/∂b = a * sigmoid(t) * (xn + c)
  - ∂/∂c = a * sigmoid(t) * b
- The vertical offset contributes a column of ones

This allows `least_squares` to use exact gradient information, improving convergence speed and reliability compared to numerical differentiation.

https://claude.ai/code/session_01VjBGnmxiob35bWRGzM8oHf


pmrv:
@prabhath-c I haven't tried it myself yet, but this should be helpful to make your fits with softplus much better.

prabhath-c:
@pmrv Thanks a lot, looks like a good idea to pre-compute the Jacobian. But unfortunately, it didn't help much with the fit. I was hoping it would either improve the speed. The issue is still the non-linear interpolation and the fit sometimes get's stuck in a local minima even after many iterations.

pmrv:
@prabhath-c Bummer, it's working though or did something break?

pmrv:
@claude If not yet in the interpolation tests, add the softplus interpolator.  There should be a hypothesis test picking random coeffs for the softplus and testing that `fit()` restores them.  Test also a piece-wise continuous function with a kink.  Softplus fit should at least recover the kink location roughly.  If this fails try to plot the loss as a function of softplus parameter and attach it to the comment so we can visually inspect the loss landscape.

pmrv:
Added `tests/unit/test_softplus.py` (commit 7f9cbe5):

- **Hypothesis test** that picks random `(a, b, c)` per softplus term plus an offset and checks that `fit()` reproduces the function on a dense grid (modulo permutation of terms).
- **Parametrized test** that a ReLU-style kink `f(x) = max(0, x - x0)` has its inflection point recovered to within ~10% of the sampling range. Passes at `x0 ∈ {-0.31, 0.0, 0.37}`.
- A couple of smoke checks (public-API export, dual interpolator role).

While writing the recovery test I tripped over the local-minimum issue @prabhath-c mentioned. The hypothesis-shrunk falsifying example is

```
y(x) = 2·softplus(2.25·x) + 2·softplus(−2·(x+1))
```

on `x ∈ [−1, 1]`. With the default `loss="soft_l1"` both amplitudes collapse to ≈ 0 and the predictor becomes nearly constant (≈ median(y)). With `loss="linear"` the fit recovers the truth to machine precision from the same initial conditions. So the Jacobian wasn't the bottleneck — the issue is the loss shape.

Loss landscape (slice at the true `(b, c, offset)`, varying the two amplitudes `(a1, a2)`):

softplus loss landscape

- Top-left: data + fits for both losses.
- Top-right: 1-D slice of the loss along `a1`. `soft_l1` is essentially flat between `a1 ≈ 0` and the true `a1 = 2`, so any gradient method launched from the default initial guess (`a ≈ amplitude/n ≈ 0.9`) drifts to the `a → 0` basin instead of climbing to the global minimum.
- Bottom row: full 2-D `(a1, a2)` slice. The `linear` landscape has a unique sharp basin at `(2, 2)`; `soft_l1` has a wide flat basin at `(0, 0)` competing with it.

I made the hypothesis test use `loss="linear"` so it actually tests recovery rather than the loss-shape problem — that way it both passes and documents the requirement.

For the underlying robustness issue, a few options that would help without changing the public API:
1. Initialize `a` closer to the true amplitude (e.g. `max(y) - min(y)` instead of `amplitude / n`) so the optimizer starts outside the `a ≈ 0` basin.
2. Multi-start from a handful of `(b, c)` perturbations and keep the best fit.
3. Stage the fit: do a short `loss="linear"` pass to get into the right basin, then refine with `soft_l1` for outlier robustness.

Happy to wire any of these in if useful — option 3 looks like the cheapest win.

---

pmrv:
Correction to my previous comment — you were right, my "basin at (0,0)" reading was wrong. The 2-D `(a1, a2)` slice I showed earlier pinned `(b, c, offset)` at the *true* values, which made it look like a single descending bowl with low loss at `(0, 0)`. That cross-section says nothing about the actual converged solution.

What `SoftplusFit(loss="soft_l1")` actually converges to from its default init on this data is

```
init  : a1=0.887  b1= 3.00  c1= 0.86 | a2=0.887  b2=-3.00  c2=-0.86 | offset=1.64
p*sl1 : a1=0.265  b1= 4.34  c1=-0.59 | a2=0.076  b2=13.69  c2= 0.33 | offset=1.29
truth : a1=2.000  b1= 1.31  c1= 0.00 | a2=2.000  b2=-1.16  c2= 1.72 | offset=0.00
```

So the amplitudes are small (0.27 and 0.08), but it's *coupled* with `b2` drifting up against the bound and the `c`s shifting, which pushes both softpluses' transition regions out of the visible window and leaves a near-constant predictor. The L1-style behaviour of `soft_l1` does drive that — when residuals are large at init (≈ 1600 in total), the per-residual IRLS weight `1/(1+r²)` is ≈ 0 exactly at the points carrying the most information (the rising right edge of `y` where the truth lives). With those residuals discounted, the surviving signal is "mostly flat", and a model with small `a` plus carefully tucked-away knees fits it.

Two cross-sections that actually show the local-minimum structure:

corrected loss landscape

- Top-right: loss along the line `p*_sl1 → truth`. Both losses *increase* before they decrease — `p*_sl1` is a genuine local minimum along this direction, the optimizer can't escape it with a Gauss–Newton step.
- Bottom-left: loss along the line `init → truth`. Monotonically downhill in both losses. So the failure isn't that init is on the wrong side of a barrier; it's that the gradient *direction* taken from init under soft_l1's IRLS reweighting curves away from that downhill path.
- Bottom-right: the residual at init and the IRLS weight. Where `|r|` is biggest (right edge), the weight is ≈ 0 — `soft_l1` is essentially saying "ignore the signal that would tell you a1, a2 should be large".

So I'd retract my "basin at (0,0)" framing. The mechanism is the IRLS reweighting that bounds large-residual influence (the "L1-like" behaviour you flagged), not a basin in the parameter cost surface. The practical fixes I suggested still apply — a smarter `a` init or a linear→soft_l1 staged fit both work because they bring init close enough to truth that all residuals are small, at which point `soft_l1` ≈ linear and the IRLS reweighting stops biasing the step.

---

pmrv:
Pushed in 2361563.

**Init fix.** Changed the initial amplitude guess from `ptp(y) / (2·n_softplus)` to `ptp(y)`. The old guess starts inside soft_l1's near-constant basin (residuals are large at every point, IRLS weights `1/(1+r²)` ≈ 0, no signal); starting from amplitudes that actually reach the data keeps residuals small enough that the IRLS reweighting isn't biased. The historical falsifying case (`2·softplus(2.25 x) + 2·softplus(-2 (x+1))`) now fits to **rms = 4e-11 in 31 ms** under the default `soft_l1`, and the hypothesis recovery test passes under the default loss without any extra knobs.

**`global_fit` method.** Wraps `scipy.optimize.differential_evolution` over an explicit box (`a ∈ [0, 5·ptp(y)]`, `b ∈ [±20]`, `c ∈ [±10]`, `offset ∈ [min(y)-ptp, max(y)+ptp]`) and polishes the DE candidate with the local fit so the result is bit-for-bit comparable to `fit()`. The DE cost reuses the same robust loss (`linear`, `soft_l1`, `huber`, `cauchy`, `arctan`) and `f_scale` as the local fit. Knobs: `seed=0` (deterministic by default), `de_maxiter=200`, `de_popsize=20`, `de_tol=1e-8`.

**`f_scale` parameter (weaker regularization).** This is the existing knob in `scipy.optimize.least_squares` — `rhõ(r²) = C²·rho(r²/C²)` with `C = f_scale` — that controls where the robust loss transitions from quadratic to linear. Default 1.0 matches the previous behaviour; **larger `f_scale` widens the quadratic region, weakening the IRLS reweighting**. New test `test_f_scale_widens_soft_l1_quadratic_region` checks that `SoftplusFit(loss="soft_l1", f_scale=1e4)` and `SoftplusFit(loss="linear")` produce predictions agreeing to 1e-3. Has no effect when `loss="linear"`.

So there are now three ways to escape the local-minimum issue, in increasing cost order:
1. Just use `fit()` — the new init handles the cases I could trigger from the existing hypothesis strategies.
2. If the local fit still gets stuck on harder data, bump `f_scale` (e.g. `f_scale=10`) to soften the IRLS reweighting while keeping outlier robustness.
3. Fall back to `global_fit()` for truly multimodal landscapes.

**Benchmark** (median of 3 reps after a warm-up, on three representative datasets):

| case | method | time | rms vs data |
|---|---|---:|---:|
| smooth alloy mixing, 121 pts (RK-shaped + linear term) | `PolyFit(4)` | 1.1 ms | 5.9e-3 |
|  | `PolyFit('auto')` | 2.9 ms | 7.2e-6 |
|  | `SGTE(4)` | 0.7 ms | 5.8e-3 |
|  | `RedlichKister(3)` | 0.5 ms | 3.6e-17 |
|  | `SoftplusFit(n=2).fit` | 66 ms | 2.8e-4 |
|  | `SoftplusFit(n=2,linear).fit` | 61 ms | 5.7e-4 |
|  | `SoftplusFit(n=2).global_fit` | **2390 ms** | 2.4e-4 |
| two softpluses (PR #82 pathological), 201 pts | `PolyFit(4)` | 0.8 ms | 1.1e-2 |
|  | `PolyFit('auto')` | 2.0 ms | 6.0e-7 |
|  | `SGTE(4)` | 0.5 ms | nan (log of negative T n/a) |
|  | `SoftplusFit(n=2).fit` | **31 ms** | **4.3e-11** |
|  | `SoftplusFit(n=2,linear).fit` | 64 ms | 3.6e-4 |
|  | `SoftplusFit(n=2).global_fit` | 2570 ms | 2.0e-4 |
| SGTE-like G(T), 80 pts | `PolyFit(4)` | 1.0 ms | 3.5e-4 |
|  | `SGTE(4)` | 1.4 ms | 5.2e-15 |
|  | `SoftplusFit(n=2).fit` | 73 ms | 6.2e-3 |
|  | `SoftplusFit(n=2).global_fit` | 2500 ms | 0.22 *(DE poorly suited to the smooth Calphad shape)* |

Headline numbers: `SoftplusFit.fit` is ~30–70× slower than the polynomial/SGTE/RK fits but typically reaches comparable or better accuracy on data that has actual softplus structure. `global_fit` is another ~40× on top of that (≈2.5 s per fit), so it's worth keeping in reserve rather than as the default.

The two-softpluses case is now the *fastest* SoftplusFit row (31 ms): once the init is right, soft_l1 converges in fewer iterations than linear does on the same data because it gives up cleanly on rounding-noise residuals.

---

pmrv:
make the benchmark scripts available in a new branch (from main) and open PR for it.

prabhath-c:
> @prabhath-c Bummer, it's working though or did something break?

@pmrv No, nothing broke. It worked essentially the same. There must be a use case where the Jacobian would help a lot. I think you can merge this change into `main`.

tests/unit/test_softplus.py
pmrv, line 1: use a common error treshold

pmrv (commented):
drop the global fit again and remove the artifacts

pmrv:
> > @prabhath-c Bummer, it's working though or did something break?
> 
> @pmrv No, nothing broke. It worked essentially the same. There must be a use case where the Jacobian would help a lot. I think you can merge this change into `main`. 

OK, it seems harmless enough to have it. I'm not sure if Claude had the same issue above as you, but did you check out the regularization and the initial guess when you had the problem before?

PR #68, opened 2026-03-11
Add ASE Thermochemistry Phase Wrappers

pmrv:
This PR integrates ASE (Atomic Simulation Environment) thermochemistry classes (`IdealGasThermo`, `HarmonicThermo`, and `CrystalThermo`) into the application as new phase objects that act like `AbstractLinePhase` objects with a fixed concentration.

Changes made:
* **Dependency Addition:** Added `ase` to the required dependencies in `pyproject.toml`.
* **New Phase Wrappers:** Created three dataclass wrappers (`ASEIdealGasPhase`, `ASEHarmonicPhase`, `ASECrystalPhase`) in `landau/phases.py`.
* **Input Handling:** Used `np.vectorize` inside the wrappers so that they can correctly process both scalar temperature inputs and NumPy arrays of temperatures (since ASE's underlying methods expect scalars).
* **Optional Imports Safety:** Wrapped the import of ASE inside a `try/except` block and created a `HAS_ASE` flag, raising a descriptive error when trying to use these phases if `ase` is not properly installed.
* **Unit Tests:** Added extensive tests checking array output shapes, scalar output types, and verifying functionality using standard ASE thermochemical inputs.

---
*PR created automatically by Jules for task [159940577703790262](https://jules.google.com/task/159940577703790262) started by @pmrv*

PR #83, opened 2026-05-04
Value-based equality and hashing for AsePhase

pmrv:
## Summary

Addresses point #2 of the Claude review on PR #68: `frozen=True` auto-generates `__hash__` from all fields including `thermochem`, but ASE's `ThermoChem` subclasses (`HarmonicThermo`, `IdealGasThermo`, ...) inherit identity-based equality from `object`. Two `AsePhase` objects built from equivalent inputs therefore compared unequal and lived in separate hash buckets.

## Implementation

`landau/phases/asewrapper.py`:

- `AsePhase.__eq__` and `__hash__` delegate to a `_key()` tuple that includes `pickle.dumps(self.thermochem)`. Pickling produces deterministic bytes that compare equal for equivalently constructed instances and naturally handles `numpy` arrays, `Atoms`, and nested `HarmonicMode` objects — without needing a bespoke recursive comparator.
- Returns `NotImplemented` for cross-type comparisons.

The `__eq__` ↔ `__hash__` contract holds: equal objects share a hash. The trade-off is that any non-pickleable attribute someone slips into a `ThermoChem` subclass would raise instead of silently degrading; this is the right failure mode.

## Tests

`tests/unit/phases/test_ase.py` adds:

- `test_ase_phase_equality_harmonic` — asserts ASE's own `HarmonicThermo` instances aren't `==`, then asserts `AsePhase` wraps them into equal objects with equal hashes.
- `test_ase_phase_equality_idealgas` — same, but for `IdealGasThermo` (exercises `Atoms` inside the pickle).
- `test_ase_phase_inequality` — flips each meaningful field one at a time and confirms inequality.
- `test_ase_phase_set_dedup` — equivalent phases collapse in a `set`.

## Notes

- The earlier `_value_equal` recursive helper (initial implementation) was replaced by `pickle.dumps` after review feedback, giving a net simpler result.
- Other review points (`np.vectorize` rebuild, `*args, **kwargs` in `__post_init__`, version pinning, etc.) are left for follow-ups.

https://claude.ai/code/session_019M5f6NrvE6HfUGsHLjh2xR

---
_Generated by [Claude Code](https://claude.ai/code/session_019M5f6NrvE6HfUGsHLjh2xR)_

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ __pycache__
 dist/
 build/
 *.egg-info
+landau/_version.py
--- a/landau/phases/asewrapper.py
+++ b/landau/phases/asewrapper.py
@@ -1,3 +1,5 @@
+import pickle
+
 import numpy as np
 from dataclasses import dataclass
 
@@ -12,6 +14,10 @@ from . import AbstractLinePhase
 class AsePhase(AbstractLinePhase):
     """
     Phase wrapper for ASE's ThermoChem classes.
+
+    Equality and hashing compare ``thermochem`` by its pickled bytes so two
+    ``AsePhase`` instances built from equivalent inputs compare equal even
+    though ASE's ``ThermoChem`` defaults to identity-based equality.
     """
     fixed_concentration: float
     thermochem: 'ThermoChem'
@@ -41,3 +47,19 @@ class AsePhase(AbstractLinePhase):
         if res.ndim == 0:
             return res.item()
         return res
+
+    def _key(self):
+        return (
+            self.name,
+            self.fixed_concentration,
+            self.pressure,
+            pickle.dumps(self.thermochem),
+        )
+
+    def __eq__(self, other):
+        if other.__class__ is not self.__class__:
+            return NotImplemented
+        return self._key() == other._key()
+
+    def __hash__(self):
+        return hash(self._key())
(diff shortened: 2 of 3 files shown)


landau/ase_phases.py
```diff
@@ -8,10 +8,39 @@
 
 from .phases import AbstractLinePhase
 
+
+def _value_equal(a, b):
```
pmrv: @claude Annoying.  Evaluate whether we can cast the thermochem into some builtin collections (to_dict/asdict/ or the like?) then rely on their equality

pmrv: Investigated. ASE's `ThermoChem` doesn't expose a `todict`/`asdict`, and the `vars()` of `HarmonicThermo` carries a list of `HarmonicMode` objects that themselves use identity equality, so any "hand off to builtin `==`" route still requires recursing into nested objects.

The simplest builtin that does survive all of that is `pickle.dumps`:

- Same inputs → identical bytes (verified for `HarmonicThermo`, `IdealGasThermo` w/ `Atoms`, `CrystalThermo` w/ numpy arrays).
- Different inputs → different bytes.
- Deterministic across repeated calls in-process.

Pushed `63f2756` replacing `_value_equal` with a single `pickle.dumps(self.thermochem)` in the equality/hash key. Net `+15 / -39` lines, same test suite passes. The trade-off is that any non-pickleable attribute someone slips into a `ThermoChem` subclass would now raise instead of silently degrading; I think that's the right failure mode but happy to revisit.

---

pmrv:
@claude fix conflicts and update  PR description.

PR #98, opened 2026-05-11
refactor: extract phase-diagram refinement into a Refiner strategy

pmrv:
## Summary

Refactors the two ad-hoc phase-diagram refinement schemes (1D scan along mu or T, Delaunay-based 2D refinement for line and triple points) into a small strategy hierarchy in a new module `landau/refine.py`. The goal is to make adding a third strategy straightforward.

### New module `landau/refine.py`

- `Refiner` ABC: each strategy implements `propose(df)` (yield candidates) and `solve(cand, phases)` (return a `RefinedPoint` or `None`). The base `.run()` handles stability filtering and column tagging.
- `RefinedPoint`: dataclass holding `(T, mu, phases)` for a located transition.
- Concrete strategies:
  - `ScanRefiner(by="mu"|"T")` — replaces `find_all_points` / `find_mu_one_point` / `find_T_one_point`.
  - `DelaunayLineRefiner` — replaces `find_triangle`.
  - `DelaunayTripleRefiner` — replaces the inline `refine_triples`.
- `default_refiners(df)` — picks the standard set based on which of T/mu are sampled.

Shared helpers (`_state_row`, `_expand_rows`, `_dominated`) replace 4 copies of the same row-building dict literal and 2 copies of the "another phase dominates" check.

### Changes to `landau/calculate.py`

- `refine_phase_diagram` becomes a thin orchestrator: split stable/unstable, add 2D border edges, run each refiner. New optional `refiners=` parameter lets callers swap or extend the defaults.
- Removed: `find_mu_one_point`, `find_T_one_point`, `find_all_points`, `find_triangle`, and the long body of the old `refine_phase_diagram`.
- `find_one_point` is re-exported from `landau.refine` for backwards compatibility (used by `tests/unit/test_calculate.py`).
- Public output columns (`stable`, `border`, `refined` with values `no`/`mu`/`T`/`delaunay`/`delaunay-triple`) are unchanged.

Net diff: +331 / -239 lines, with `calculate.py` shrinking from 460 to ~220 lines.

### Adding a new strategy

```python
class MyRefiner(Refiner):
    label = "my-new"
    def propose(self, df): ...
    def solve(self, cand, phases): ...

calc_phase_diagram(..., refine=True)  # default behavior unchanged
# or
refine_phase_diagram(df, phases, refiners=[..., MyRefiner()])
```

## Test plan

- [x] `pytest tests/` — 68 passed, 2 skipped (same as baseline before refactor)
- [x] Spot-check a 2D phase diagram visually with `landau.plot.plot_phase_diagram` to confirm the Delaunay refinement still produces identical border points
- [x] Confirm downstream `landau.poly` segment construction still works against the `refined` column

https://claude.ai/code/session_017jy31tSCWgfm2Z2tDk3FBe

---
_Generated by [Claude Code](https://claude.ai/code/session_017jy31tSCWgfm2Z2tDk3FBe)_

### DIFF
--- a/landau/calculate.py
+++ b/landau/calculate.py
@@ -2,20 +2,17 @@
 Calculates phase diagrams from sets of Phases.
 """
 
-from functools import partial
 import numbers
-import warnings
-from typing import Iterable
+from typing import Iterable, Sequence
 
 import numpy as np
 import pandas as pd
-import scipy.optimize as so
-from scipy.spatial import Delaunay
 from scipy.constants import Boltzmann, eV
 from sklearn.cluster import AgglomerativeClustering
 
 
 from .phases import Phase, AbstractLinePhase
+from .refine import Refiner, default_refiners, _find_one_point as find_one_point  # noqa: F401
 
 
 kB = Boltzmann / eV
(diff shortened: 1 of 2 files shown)


pmrv:
give solve the option to return more than one refinedpoint.

PR #111, opened 2026-05-12
Add ASE Thermochemistry Phase Wrappers (squashed #68)

pmrv:
Squashed replacement for #68, rebased onto current `main`.

Integrates ASE (Atomic Simulation Environment) thermochemistry classes (`IdealGasThermo`, `HarmonicThermo`, `CrystalThermo`) as phase wrappers that behave like `AbstractLinePhase` objects with a fixed concentration.

## Changes

- **Dependency:** Add optional `ase` extra in `pyproject.toml` (also pulled into the `test` extra).
- **Subpackage split:** `landau/phases.py` → `landau/phases/__init__.py`, with a new `landau/phases/asewrapper.py` providing a unified `AsePhase` dataclass.
- **Vectorization:** `semigrand_potential` is vectorized over temperature, so scalar and array inputs both work even though ASE's underlying methods expect scalars.
- **Optional import safety:** ASE is imported behind a `pyiron_snippets` `ImportAlarm`, so the package remains importable without `ase` installed.
- **Tests:** New `tests/unit/phases/test_ase.py` and `tests/unit/phases/test_vectorize.py`; removes the now-obsolete `tests/unit/test_phases_utils.py`.

## Relation to #68

Closes #68 (replaced by this PR). All 18 commits from the original branch are squashed into a single commit on top of current `main`. The companion notebooks PR #85 will be retargeted onto this branch.

---


landau/phases/asewrapper.py
```diff
@@ -0,0 +1,65 @@
...
+
+from pyiron_snippets.import_alarm import ImportAlarm
+
+with ImportAlarm("ASE is required to use ASE phase wrappers. Install with pip install 'landau[ase]'") as ase_alarm:
+    from ase.thermochemistry import ThermoChem
+
+from . import AbstractLinePhase
+
+@dataclass(frozen=True)
+class AsePhase(AbstractLinePhase):
```
pmrv, line 14: The distinction to call gibbs or helmholtz methods based on pressure won't actually work, since ASE classes not always define both.  Read the original docs again and come up with some strategies.  Consider that landau does not yet understand pressures, but might in the future.

pmrv: Good catch. Fixed in 026ae02 by switching to capability-based dispatch:

- Prefer `get_helmholtz_energy` when the `ThermoChem` defines it (HarmonicThermo, CrystalThermo, HinderedThermo). landau has no pressure concept yet, so this is the default path.
- Fall back to `get_gibbs_energy(T, pressure=self.pressure)` when only the Gibbs method exists (IdealGasThermo). If `pressure` is `None`, default to 1 atm so the call doesn't fail; a real `pressure` field is kept on the dataclass for when landau learns about pressure.
- Raise `TypeError` if a `ThermoChem` subclass exposes neither method, so the failure is loud rather than an `AttributeError` deep in vectorized code.

When landau grows pressure-as-a-thermodynamic-variable, the natural extension is to make `line_free_energy` take `(T, p)` and route both Helmholtz and Gibbs paths through it (Helmholtz ignores `p`, Gibbs uses it), at which point the stored `pressure` field can become a default rather than the only option.

---

tests/unit/phases/test_ase.py
```diff
@@ -0,0 +1,89 @@
...
+    assert ht1 is not ht2
+    assert ht1 != ht2
+
+    p1 = AsePhase("ht", 0.3, thermochem=ht1)
+    p2 = AsePhase("ht", 0.3, thermochem=ht2)
+    assert p1 == p2
+    assert hash(p1) == hash(p2)
+
+
+@pytest.mark.skipif(ase_alarm.message is not None, reason="ASE is not installed")
```
pmrv: there's a module wide skip marker you can use

landau/phases/asewrapper.py
pmrv: Done in `cd41012`. Replaced the six per-test `@pytest.mark.skipif` decorators with a single `pytestmark = pytest.mark.skipif(...)` at module scope; also dropped the unused `hypothesis` imports while I was in there.

---

tests/unit/phases/test_ase.py
pmrv: Done in `cd41012`. Replaced the six per-test `@pytest.mark.skipif` decorators with a single `pytestmark = pytest.mark.skipif(...)` at module scope; also dropped the unused `hypothesis` imports while I was in there.

---

PR #112, opened 2026-05-12
ASE notebooks: drop EMT-Fe section, add Cu–Ag miscibility example

pmrv:
Replaces #85 (whose branch is no longer pushable from this session). Layered on top of #111.

### `notebooks/ASE/FCC_BCC_Fe.ipynb` — drop EMT phonon section
At pmrv's request, the EMT phonon section (Ni at Fe lattice constants) is removed. The Debye model alone now carries the FCC↔BCC example; sections are renumbered. The notebook is re-executed end to end (5/5 code cells, 0 errors).

### `notebooks/ASE/EMT_CuAg.ipynb` — new alloy example
Replacement EMT-phonon example using Cu–Ag, where EMT actually predicts a positive heat of mixing (~32 meV/atom at x=0.5):

1. Build five FCC supercells along the Cu–Ag isomorph (x_Cu ∈ {0, 0.25, 0.5, 0.75, 1}) with Vegard lattice constants.
2. Static EMT energies → mixing-enthalpy curve.
3. Phonon DOS at each composition via `ase.phonons.Phonons` + EMT.
4. Wrap each sampled composition as `AsePhase(thermochem=CrystalThermo(...))`.
5. Combine into a `RegularSolution(add_entropy=True)` and reconstruct the binodal of the resulting miscibility gap from the lower convex hull of `F(c)` at each `T` (the existing `plot_phase_diagram` hits a pre-existing pandas-3 bug in `cluster_phase`, so the binodal is built explicitly).

EMT predicts a critical T ≈ 629 K — much lower than the experimental Cu–Ag eutectic (~1052 K), but the workflow generalises to any `ase.thermochemistry` model. Notebook executes 6/6 cells, 0 errors.

### `notebooks/ASE/Hydrate.ipynb` — fix import
`landau.ase_phases` → `landau.phases.asewrapper` after the subpackage split. Re-executed (12/12 cells, 0 errors).

### `.gitignore`
Add `notebooks/ASE/cuag-*-phonon/` for the EMT phonon caches.

## Why a new branch

PR #85's branch (`notebooks-ase-phases`) is no longer pushable from this session — the session proxy has restricted it to PR #111's branch only. This branch is the same content rebased onto #111.

## Test plan

- [x] All three notebooks fully executed end-to-end (no error outputs)
- [ ] Visual sanity check of the Cu–Ag binodal plot
- [ ] Decide whether to also patch `landau/plot.py:cluster_phase` for pandas 3 (out of scope for this PR)

---


pmrv:
expound on what the 'pandas 3' bug is and fix it in a separate PR.

pmrv:
When constructing the disordered supercells relax the atomic positions before calculating phonons.

pmrv:
Bug write-up + fix in #113.

**The bug.** In pandas 3, `df.groupby('phase').apply(cluster, include_groups=False)` no longer concatenates per-group results into a single Series when `cluster` returns a `pd.Series` whose index is the *original* row indices (not a fresh `RangeIndex`). Instead the apply result becomes a one-row DataFrame indexed by the group key, with columns equal to the original row indices. The very next statement in `cluster_phase` —

```python
df["phase_unit"] = df.groupby("phase", group_keys=False).apply(cluster, ...)
```

— then raises `ValueError: Cannot set a DataFrame with multiple columns to the single column phase_unit`. Pandas 2 unwrapped this to a Series and the assignment worked.

This made `plot_phase_diagram` unusable under pandas 3 for any input that reaches `cluster_phase` — which is every non-trivial diagram, including the Cu–Ag miscibility example here.

**The fix** in #113 replaces the `apply` with an explicit per-group loop that builds a row-aligned Series, equivalent to the old behavior under pandas 2. Regression test added.

The other two `groupby(...).apply(plot_tie, include_groups=False)` sites in `plot.py` are side-effect calls whose return value is discarded, so they're unaffected.

---

pmrv:
Done in e550cd2. Switched the disordered samples to 32-atom 2×2×2 conventional FCC supercells with random Cu placements (fixed RNG seed). The previous 4-atom ordered cells sat at force-zero sites by symmetry so BFGS would have been a no-op — the relaxation only does something on configurations that actually break the on-site symmetry.

Effect of the relaxation:

| x_Cu | E_mix unrelaxed (meV/atom) | E_mix relaxed (meV/atom) |
|------|-----------------------------|---------------------------|
| 0.25 | 42.0 | 28.4 |
| 0.50 | 62.8 | 40.4 |
| 0.75 | 37.8 | 25.8 |

Critical T of the miscibility gap moves from ~630 K (ordered, unrelaxed) to ~1000 K (random, relaxed) — much closer to the experimental Cu–Ag eutectic at ~1052 K. Phonons run on the relaxed cells; total notebook execution stays under a minute. The mixing-enthalpy figure now shows both unrelaxed and relaxed curves so the effect is visible.

---

pmrv:
rebase relevant commits again.  plotting finally fixes you can just call the normal plotting routines now

pmrv:
fix deprecation warnings where possible,  ensure all plots are rendered with tsp dependencies installed.  CuAg makes some weird physical statements, double check everything.  Rather say nothing than something wrong.

notebooks/ASE/Hydrate.ipynb
pmrv, line 1: Change to Mg-H with some H solubility in the Mg phase

For the intermetallic MgH2, use the energy scraped from here

https://next-gen.materialsproject.org/materials/mp-23710

and tune the entropy to reproduce the melting point here

https://next-gen.materialsproject.org/contribs/contributions/65cfa8cd1db79045ca68a428

Try and use the melting and evaporation temperatures of Mg to tune the entropy of the liquid and solid.

Make it otherwise very clear that this notebook is for illustration only and numbers are note reliable.

pmrv: Keep only one cell with the final phase diagram (using the default poly method).

pmrv: increase temperature range to include liquid -> gas transition across all concentrations.

notebooks/ASE/EMT_CuAg.ipynb
pmrv, line 1: The cell printing the lattice parameter as a function of conc doesn't take super cell size into account.

pmrv: phonon DOS seems oversmeared.

pmrv: do not mention MiscibilityGapRefiner, that's an implementation detail.

pmrv: Ensure that the entire notebook reads reasonable to a first time reader that hasn't followed the conversion here.  No references to ahistorical review feedback.

notebooks/ASE/FCC_BCC_Fe.ipynb
pmrv, line 1: EMT is no longer part of the notebook, do not mention

pmrv (commented):
revise, ensure each notebook reads clean, makes no reference to its history or review feeback.

notebooks/ASE/EMT_CuAg.ipynb
pmrv, line 1: This part

```python
compositions = [0.0, 0.25, 0.5, 0.75, 1.0]
cells = {x: make_cell(x) for x in compositions}
for x, c in cells.items():
    n_cu = sum(s == 'Cu' for s in c.get_chemical_symbols())
    print(f'x_Cu = {x:.2f}  N = {len(c):2d}  ({n_cu} Cu, {len(c)-n_cu} Ag)  a = {c.cell.cellpar()[0]:.3f} Å')
```
needs to define cellpar by the supercell number.

notebooks/ASE/FCC_BCC_Fe.ipynb
pmrv, line 1: some unicode like this ' \u2192' doesn't render properly.

notebooks/ASE/Hydrate.ipynb
pmrv, line 1: Materials Project predict 690K for MgH2 melting point, not 560.

pmrv, line 1: In the sanity check section, use calc_phase_diagram in the 1d case as is done in other notebooks.

pmrv, line 1: try a finer mesh for the 2d phase diagram to avoid the final warnings.

pmrv (commented):
revise.

notebooks/ASE/Hydrate.ipynb
pmrv, line 1: 2D phase diagram needs a wide mu range

PR #113, opened 2026-05-12
Fix cluster_phase under pandas 3

pmrv:
## What

In pandas 3, `df.groupby('phase').apply(cluster, include_groups=False)` no longer concatenates per-group results into a single Series when `cluster` returns a Series indexed by the original row indices. Instead the result becomes a one-row DataFrame indexed by the group key, with columns equal to the original row indices. The next line — `df["phase_unit"] = <that DataFrame>` — then raises:

```
ValueError: Cannot set a DataFrame with multiple columns to the single column phase_unit
```

This was reachable from any `plot_phase_diagram` call under pandas 3 — including the Cu–Ag miscibility-gap example added in #112, which is what surfaced it.

## Fix

Replace the `apply` with an explicit per-group loop that returns a row-aligned Series. This is portable across pandas 2 and 3 and equivalent in behavior to the old code on pandas 2.

```python
parts = [cluster(g, use_mu=False) for _, g in df.groupby("phase", sort=False)]
df["phase_unit"] = pd.concat(parts).reindex(df.index)
```

## Why this only

The other two `groupby(...).apply(plot_tie, include_groups=False)` sites in `plot.py` (lines 129, 147) call `plot_tie` for matplotlib side effects only and discard the return value, so the wrapping change does not affect them.

## Test

`tests/regression/test_cluster_phase_pandas3.py` exercises both single-phase and two-phase inputs and asserts the resulting `phase_unit`/`phase_id` columns are well-formed. Both pass on pandas 3.0.3.

## Out of scope

Anything else in `plot.py` / `cluster_phase`'s clustering parameters; this PR only restores pandas-3 compatibility.

---

### DIFF
--- a/landau/plot.py
+++ b/landau/plot.py
@@ -6,7 +6,6 @@ from pyiron_snippets.deprecate import deprecate
 import matplotlib.pyplot as plt
 import seaborn as sns
 import numpy as np
-import pandas as pd
 
 from .calculate import get_transitions, cluster
 import landau.poly as poly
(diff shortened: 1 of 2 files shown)


pmrv:
I'm not entirely sure I believe you.  Read [the docs](https://pandas.pydata.org/docs/user_guide/groupby.html#flexible-apply) and get back to me.  There must be a more aesthetically pleasing way than this.  Also keep in mind pandas 2/3 cross compat.  I won't turn off pandas 2 support just yet.

pmrv:
Fair — went back to the [Flexible apply](https://pandas.pydata.org/docs/user_guide/groupby.html#flexible-apply) docs and you're right that there's a tidier path. The relevant rule there is that `apply` combines DataFrame return values consistently along the row axis regardless of group count, while Series return values are the inconsistent case.

What I had missed digging into earlier is that the pandas 3 bug only fires with a **single** group — multi-group inputs concatenate fine. I had been reading "all Series returns are now broken" but it's specifically the single-group case that gets wrapped into a 1×N DataFrame indexed by the group key, which is why this slipped through everywhere except the Cu-Ag miscibility-gap example (one `RegularSolution` phase).

Pushed 4c6aacb: drops the explicit per-group loop and instead wraps the callable's return in `.to_frame("phase_unit")`. apply now sees a DataFrame from the callable, concatenates it the same way in pandas 2 and 3, and the assignment picks the column back off:

```python
df["phase_unit"] = df.groupby("phase", group_keys=False).apply(
    lambda g: cluster(g, use_mu=False).to_frame("phase_unit"),
    include_groups=False,
)["phase_unit"]
```

Regression test (both single- and two-phase inputs) still passes; `import pandas as pd` removed since it's no longer needed.

---

PR #115, opened 2026-05-12
Add ClausiusClapeyronRefiner for tracing coexistence lines

pmrv:
## Summary

Adds a new `ClausiusClapeyronRefiner` in `landau/refine.py` that adapts the predictor-corrector tracer from `scripts/cc_demo.py` into the existing `Refiner` strategy interface.

- **Propose**: walks the (mu, T) Delaunay tessellation and seeds one trace per two-phase simplex, deduped by unordered phase pair.
- **Solve**: from each seed, walks both directions along T using the finite-difference Clausius-Clapeyron slope `dmu*/dT` as a linear predictor and a bracketed bisection on `phi1(T,mu) - phi2(T,mu)` as the corrector. Step size adapts to the local slope so the bracket stays valid.
- Emits a dense sequence of `RefinedPoint`s per seed (vs. one point per simplex for `DelaunayLineRefiner`), giving much finer sampling of coexistence lines for the same input grid.

Exposed via `landau.refine.__all__`; not added to `default_refiners` (opt-in for now).

## Test plan

- [x] `pytest tests/unit/test_refine.py` — 4 new tests covering tracing, pair dedup, label, and the no-coexistence case all pass.
- [x] Verified the refiner produces transition points where `|phi_A - phi_B| < 1e-7` across the T range of a two-phase LinePhase fixture.
- [x] Full unit suite still passes (one pre-existing PolyFit hypothesis failure, unrelated).

https://claude.ai/code/session_01QBjNvugbRiwHDsKLrgX9vM

---
_Generated by [Claude Code](https://claude.ai/code/session_01QBjNvugbRiwHDsKLrgX9vM)_


landau/refine.py
```diff
@@ -267,6 +268,165 @@ def triplemin(x):
...
+class _CCCandidate:
+    phase1: str
+    phase2: str
+    T_seed: float
+    mu_bracket: tuple[float, float]
+    T_min: float
+    T_max: float
+
+
+def _refine_isothermal_jump(phase1, phase2, T, mu_lo, mu_hi, tol=1e-9):
```
pmrv: Is this different from the existing find_one_point? If not reuse

landau/refine.py
```diff
@@ -267,6 +268,165 @@ def triplemin(x):
...
+    max_steps : int
+        Hard cap on steps per direction to bound work if the bracket
+        breaks down.
+    """
+
+    label = "clausius-clapeyron"
+
+    def __init__(self, dT_max: float = 50.0, tol: float = 1e-9,
+                 max_steps: int = 500):
+        self.dT_max = dT_max
```
pmrv, line 579: From what I saw in the demo plot it actually looks as if we're getting a bit too fine just below T_c.  Think again about the critical scaling and at the very least introduce a minimum temperature step to avoid wasting time.

landau/refine.py
```diff
@@ -267,6 +268,165 @@ def triplemin(x):
...
+                    mu_predicted - bracket, mu_predicted + bracket,
+                    tol=self.tol,
+                )
+            except ValueError:
+                return
+            dmu_dT = (mu_next - mu_star) / dT
+            T, mu_star = T_next, mu_next
+            yield RefinedPoint(T=T, mu=mu_star, phases=(name1, name2))
+
+    def solve(self, cand: _CCCandidate, phases) -> list[RefinedPoint]:
```
pmrv: I want some kind of provision to avoid retracing the same line multiple times.  One could be more clever, but at the very least maintain a lean list of traces seen for a phase pair and if the simplex straddles a previously seen line, skip it.

pmrv (commented):
reivse

pmrv:
add to default refiners if 2d data is available

pmrv:
Write a new notebook that tests this one a simple regular solution phase with repulsive interaction parameter.

landau/refine.py
```diff
@@ -267,6 +268,248 @@ def triplemin(x):
         return [RefinedPoint(T=T, mu=mu, phases=names)]
 
 
+# -- Clausius-Clapeyron tracing refiner --------------------------------------
+
+
+@dataclass(frozen=True)
+class _CCCandidate:
```
pmrv: needs docs

landau/refine.py
```diff
@@ -267,6 +268,320 @@ def triplemin(x):
...
+    T_seed: float
+    mu_bracket: tuple[float, float]
+    T_bracket: tuple[float, float]
+    T_min: float
+    T_max: float
+    existing: tuple[tuple[float, float], ...] = field(default=())
+    same_phase: bool = False
+
+
+class ClausiusClapeyronRefiner(Refiner):
```
pmrv: update docs for same phase case

landau/refine.py
```diff
@@ -322,59 +323,135 @@ class ClausiusClapeyronRefiner(Refiner):
...
+            elif n == 1:
+                # Single-phase simplex straddling the binodal midpoint
+                # signals an internal miscibility gap: vertices on each
+                # side carry the two coexistence concentrations.
+                c_min = float(simplex["c"].min())
+                c_max = float(simplex["c"].max())
+                if (c_max - c_min < self.c_jump_min
+                        or c_min > self.c_midpoint
+                        or c_max < self.c_midpoint):
+                    continue
```
pmrv: get rid of midpoint, we do not know this in general

notebooks/ClausiusClapeyron.ipynb
pmrv, line 1: the comparison to delauney makes no sense, since that one doesn't handle intraphase demixing.
Results from cc_pts after refining are actually terrible, see plot below.  I think you are fudging the second graph in the notebook.  You are using the analytic concentrations at the cc refined temperatures for the second plot, thats wrong.

Image

pmrv (commented):
revise

pmrv:
outputs are extremely noisy and useless

image

something must be wrong with the reference phase, look at this plot of the coarse data

image

pmrv:
tracer does not stop on its own once the miscibility gap closes, but it should
by leaving the temperature range below T_c you are cheating

image

Another issue is that Refine will now evaluate the same phase on the same mu/T leading to the same filled in value for conc.  The refiner you build traces the same coexistance line but does not (as do the others) add two distinct points for each phase.  We'll need to fudge it in that the refiner base class shifts the mu a tiny bit left/right before evaluating conc.  It is however crucial for the rest of the package that the mu points of the two refined resulting points are exactly equal.  Let's do this for now.  Emit a RefinedMiscibilityGap with one phase inside and teach Refiner to do what I outlined above.

pmrv:
Still not working.

image


Just for the record:
1. still overshooting
2. temperature step too large near T_c now.  seems constant actually

For now concentrate on the duplicate detection.  It still seems that the whole thing is traced multiple times.

For debugging make the row expander and solver keep info on which starting seed generated the refined points.

Expand also the notebook by adding a section where you build all the steps of how to call the Refiner, so I can follow along easily to debug.

pmrv:
image
already much better.

next the problem of temperature steps:  In the notebooks this is the difference observed
```
array([ 6.81868858,  1.        ,  1.        , 50.        , 50.        ,
       50.        , 50.        , 50.        , 50.        , 50.        ,
       50.        , 50.        , 50.        , 50.        ])
```

so it seems to increase and hit one of your limits, probably you're misscaling something somewhere.

pmrv:
image
we're getting there but you picked up a drift and are overshooting again, though it looks again like this could be something to do with the reference free energy

pmrv:
write some tests that applies the new refiner on an interphase boundary too, pick any example from the existing notebooks that's simple.  add it to my debug notebook too

pmrv:
now split the refiner in two classes for inter and intra phase boundaries.

pmrv:
Notebook outputs for the current state of the branch:

**1. Free energy curves $f(T,c)$** — analytical regular solution at four temperatures, double-well below $T_c$ ≈ 580 K, single minimum above.

free energy curves

**2. Miscibility-gap result** — `MiscibilityGapRefiner` on the analytical regular solution. Left: coexistence chemical potential $\mu^*(T) \equiv 0$ by symmetry. Right: refined points (red) on the analytical binodal (black). Trace stops at $T = 573$ K, ~7 K below $T_c$.

miscibility gap

**3. Walkthrough plot — refined points coloured by seed** — single seed traces the whole line, no retraces.

seed colouring

**4. Inter-phase boundary** — `ClausiusClapeyronRefiner` on the solid/liquid ideal-solution binary lifted from `IdealSolution.ipynb`. Left: coarse phase samples in $(\mu, T)$ plus refined coexistence coloured by seed. Right: resulting solidus / liquidus.

interphase


---

pmrv:
does the interphase case show some retracing??

pmrv:
take a refactor and simplification pass

pmrv:
I also want to see a test with an asymmetric miscibility gap,  use the SlowInterpolatingPhase instead of RegularSolution this time.

pmrv:
Updated notebook output after the refactor and the new asymmetric-gap section:

**1. Free energy curves $f(T, c)$** — analytical regular solution, double well below $T_c \approx 580$ K.

free energy curves

**2. Symmetric miscibility-gap result** — `MiscibilityGapRefiner` on the analytical (symmetric) regular solution. $\mu^*(T) \equiv 0$, trace auto-stops at $T = 573$ K.

symmetric gap

**3. Single-seed trace** — walkthrough plot coloured by seed; no retracing.

seed colouring

**4. Inter-phase boundary** — `ClausiusClapeyronRefiner` on the solid/liquid ideal-solution binary. With the single-point dedup fix, the four lone seeds have collapsed to three (seed 1 is now blocked by seed 0).

interphase

**5. Asymmetric miscibility gap (new)** — sub-regular solution $f_\text{mix} = c(1-c)(L_0 + L_1(2c-1))$ with $L_0=0.10, L_1=0.04$ eV via `SlowInterpolatingPhase` + `RedlichKister(2)`. $\mu^*(T)$ drifts from 0 (median $\approx 15$ meV) and the binodal sits skewed off the symmetric axis $c=0.5$ ($c_\text{left} + c_\text{right} \approx 1.27$).

asymmetric gap


---

landau/refine.py
```diff
@@ -76,6 +85,21 @@ class RefinedMiscibilityGap:
...
+    def to_rows(self, phases: Mapping[str, Phase]) -> list[dict]:
+        ph = phases[self.phase]
+        rows = []
+        for probe_mu in (self.mu_left, self.mu_right):
+            row = _state_row(ph, self.T, probe_mu)
+            # Force the recorded mu to be the exact coexistence value
+            # so both branch rows share the same (T, mu).
+            row["mu"] = self.mu
+            row["seed"] = self.seed
+            rows.append(row)
```
pmrv: we keep this behavior for now but make a note in a new issue already that in the future I want the refiner to tag the refined points explicitly as belonging to the same phase boundary.  It's kind of ridiculous that we have to resort as to grouping on random numeric values right to reextract that information during plotting when we already have it here.

landau/refine.py
pmrv, line 1: I think we can drop seed attribute on the candidates again, since they were just for debugging.

landau/refine.py
```diff
@@ -267,6 +328,487 @@ def triplemin(x):
...
+                   tol_mu: float) -> bool:
+    """Is (T, mu) within ``tol_mu`` of the piecewise-linear ``line``
+    (T-sorted) at the same T?"""
+    if not line or T < line[0][0] or T > line[-1][0]:
+        return False
+    for (T1, mu1), (T2, mu2) in zip(line, line[1:]):
+        if T1 <= T <= T2 and T2 > T1:
+            mu_interp = mu1 + (mu2 - mu1) * (T - T1) / (T2 - T1)
+            if abs(mu - mu_interp) <= tol_mu:
+                return True
```
pmrv: surely this can be simplified by using scipy.interpolate for a piecewise linear interp?  Try it and keep if it looks simpler, if not tell what didn't work

landau/refine.py
```diff
@@ -267,6 +328,487 @@ def triplemin(x):
...
+    return False
+
+
+def _simplex_straddles(cand, traced: list[tuple[float, float]]) -> bool:
```
pmrv: we already depend on shapely, could it make this function simpler?

landau/refine.py
```diff
@@ -267,6 +328,487 @@ def triplemin(x):
...
+    def _dT_adapt(self, step: _StepResult, dmu_dT: float,
+                  half_width: float) -> float:
+        """Step size for the next predictor step. Default keeps the
+        predicted mu drift bounded by ``half_width``."""
+        slope = max(abs(dmu_dT), 1e-9)
+        return half_width / slope
+
+    # -- shared trace skeleton ---------------------------------------------
+
+    def _trace(self, cand, phases, T0, mu0, half_width, T_target, sign):
```
pmrv, line 703: needs some friendlier docstrings and explanation

landau/refine.py
```diff
@@ -267,6 +328,487 @@ def triplemin(x):
...
+        """Walk from (T0, mu0) toward T_target; yield refined transitions."""
+        dT_boot = sign * min(self.dT_min, abs(T_target - T0))
+        if dT_boot == 0:
+            return
+        T_b = T0 + dT_boot
+        try:
+            step = self._refine_step(
+                cand, phases, T_b, mu0 - half_width, mu0 + half_width)
+        except ValueError:
+            return
```
pmrv, line 760: IS this really exercised?  If so it should probably warrant a warning and/or a bug fix.

landau/refine.py
```diff
@@ -267,6 +328,487 @@ def triplemin(x):
...
+    ----------
+    dT_max, dT_min, max_steps :
+        See :class:`_CCBase`.
+    c_jump_min : float
+        Minimum c-spread of a simplex's vertices for it to be seeded.
+    gap_close : float
+        Trace-time absolute c-jump threshold (default 1e-3).
+    gap_share_min : float
+        Dominant-segment share of total scan c-variation below which
+        the bracket is judged supercritical (default 0.5).
```
pmrv: default should be .1 as we discussed before

pmrv (commented):
revise

landau/refine.py
```diff
@@ -267,6 +313,488 @@ def triplemin(x):
         return [RefinedPoint(T=T, mu=mu, phases=names)]
 
 
+# -- Clausius-Clapeyron tracing refiner --------------------------------------
+
+
+# -- Shared trace helpers (used by both Clausius-Clapeyron refiners) ----------
+
+
+def _point_on_line(T: float, mu: float,
```
pmrv, line 478: check whether would get even simpler with shapely

landau/refine.py
pmrv, line 1: docstrings, document *all* arguments

landau/refine.py
```diff
@@ -267,6 +313,488 @@ def triplemin(x):
...
+            if sign * (T_target - T) <= 1e-12:
+                return
+            dT_adapt = self._dT_adapt(step, dmu_dT, half_width)
+            dT_adapt = min(self.dT_max, max(self.dT_min, dT_adapt))
+            dT = sign * dT_adapt
+            if sign * (T + dT - T_target) > 0:
+                dT = T_target - T
+
+            T_next = T + dT
+            mu_predicted = mu_star + dmu_dT * dT
```
pmrv: break that out into a little method, so we may put better predictors here in the future.

pmrv (commented):
revise

pmrv:
remove the delaunay pair refiner from the default refiner list, keep just the triple point one.

pmrv:
squash pr

landau/refine.py
```diff
@@ -267,15 +409,710 @@ def triplemin(x):
...
+def _trace_geom(points: list[tuple[float, float]]):
+    """One trace -> one shapely geometry. ``Point`` for length 1,
+    ``LineString`` otherwise."""
+    if len(points) == 1:
+        return shapely.Point(*points[0])
+    return shapely.LineString(points)
+
+
+def _point_on_line(T: float, mu: float,
+                   traces: tuple[tuple[tuple[float, float], ...], ...],
```
pmrv: Invent some cute type aliases to make this easier to read across the module

PR #110, opened 2026-05-12
refactor: remove scalar-vs-array branching in three places

pmrv:
Removes the `isinstance`/`ndim` dispatch pattern in three spots, replacing with uniform numpy idioms (`np.where`, `np.full`, `.item()`, direct arithmetic broadcasting).

Closes #105

Generated with [Claude Code](https://claude.ai/code)

Issue #117, opened 2026-05-12
Refactor: extract axis-setup from `_plot_phase_diagram`


pmrv:
In `landau/plot.py` the helper `_plot_phase_diagram` interleaves three responsibilities (lines 157-204):

1. Building polygons and adding them to the axis (the actual plotting).
2. Optional tieline overlay.
3. Axis labelling / xlim selection based on `variables[0]`:

```python
if variables[0] == "c":
    ax.set_xlim(0, 1)
    if element is not None:
        ax.set_xlabel(rf"$c_\mathrm{{{element}}}$")
    else:
        ax.set_xlabel("$c$")
elif variables[0] == "mu":
    mus = df_stable["mu"].unique()
    mus = mus[np.isfinite(mus)]
    if len(mus) > 0:
        ax.set_xlim(mus.min(), mus.max())
    ...
else:
    raise ValueError(...)
```

This is the only spot in the plotting path that is sensitive to the choice of x-axis variable, and it's currently untested. Extracting a `_set_axis_for(variables[0], ...)` (or a small dispatch table keyed on the axis name) gets us:

- A pure function that can be unit-tested with a stub `Axes` (or even just a `MagicMock`) — no figure, no polygons, no calc pipeline.
- A natural place to add ternary-diagram support later (related to #34, #17).
- A clean separation: `_plot_phase_diagram` becomes "compute polygons, plot them, maybe overlay tielines, then set up axis".

## Scope

- Touch only `landau/plot.py`.
- Add unit tests in `tests/unit/plot/` exercising:
  - `variables=["c", ...]` with and without `element`
  - `variables=["mu", ...]` with finite mus, all-non-finite mus, and `element`
  - Unknown axis name raises `ValueError`

## Out of scope

- Generalising tieline plotting to mu-space (still only triggered for `variables[0] == "c"`).
- Anything in `plot_1d_*_phase_diagram`.

Part of #116.

pmrv:
@claude go

PR #120, opened 2026-05-12
test: cover get_polygons and plot_polygons (closes #88)

### LINKED ISSUE #88: Add test coverage for get_polygons and plot_polygons
## Problem

`get_polygons` and `plot_polygons` are new public API functions introduced by PR #57, but they have no test coverage. The test file `tests/unit/test_issue_34.py` that was written for this purpose was dropped in commit `ddd8654` with the message "Drop unit test; not well understood enough".

## Impact

These functions are exported as part of the public API (or should be — see related issue about `__all__`), but their contract is unverified. Regressions in polygon generation or plotting will not be caught by CI.

## Suggested test contract

- `get_polygons` should return the same polygon geometries as the original pre-refactor `plot_phase_diagram` / `plot_mu_phase_diagram` (embed the original implementations in the test suite for comparison, as suggested in PR review).
- `plot_polygons` should add the expected number of `matplotlib.patches.Polygon` artists to the provided `Axes`.
- Tests should be parametrized over all relevant phase classes (using `@pytest.mark.parametrize` or Hypothesis strategies).

Identified during review of PR #57.

pmrv:
Adds 14 small, orthogonal unit tests for the public polygon-plotting plumbing in `landau/plot.py`. Closes #88.

## Why

`get_polygons` and `plot_polygons` are public-API since PR #57, but the dedicated test file `tests/unit/test_issue_34.py` was removed in commit `ddd8654` ("Drop unit test; not well understood enough") and never replaced. Both functions are now exported and used by `plot_phase_diagram`, but regressions are only caught indirectly through the higher-level plotting tests.

## What

A new `tests/unit/plot/test_polygons.py` with 14 tests. Each test exercises exactly one behaviour; they share two tiny fixtures (`_stable_df` for synthetic input, `_polys_series` for fake polygon collections).

**`get_polygons` (7 tests):**
- Filters non-stable rows.
- Does not mutate the caller's DataFrame.
- Defaults `variables` to `["c", "T"]`.
- Forwards explicit `variables` through to the poly method.
- Forwards `**kwargs` (`alpha`, `min_c_width`) to `handle_poly_method`.
- Drops `phase_unit == -1` rows and emits a `UserWarning`.
- Returns a `pd.Series` of `matplotlib.patches.Polygon`, indexed by `(phase, phase_unit)`.

**`plot_polygons` (7 tests):**
- Adds one patch per item to the supplied Axes.
- Uses the supplied Axes (not `plt.gca`) when one is given.
- Falls back to `plt.gca` when `ax=None`.
- Looks up color from `color_map` by phase name (scalar key).
- For `(phase, rep)` multi-index keys, appends `rep` apostrophes to the label.
- Always sets edgecolor to black.
- Assigns zorder inversely proportional to polygon bbox area.

The forwarding tests use a small `_CapturingPoly` stub (a real subclass of `AbstractPolyMethod` so it survives `handle_poly_method`'s type check) to keep them independent of the actual polygon construction. The smoke test that pins the return type uses `Concave(drop_interior=False)` because the synthetic frame has no `border` column.

## Test plan

- [x] `pytest tests/unit/plot/test_polygons.py -v` → all 14 pass locally.
- [x] Full suite (excluding the ASE-only files which fail to collect without ASE) is green: 110 passed, 6 pre-existing ASE failures unchanged.

## Notes

- I avoided touching `landau/plot.py` itself — this PR is pure test coverage, by design. There are documented follow-up refactors in #116 (axis-setup extraction, DRY of `phase_id`/`border_segment` formatting, splitting `cluster`).
- One incidental finding while writing the tests: when `AbstractPolyMethod.apply`'s `groupby(...).apply(...)` produces an empty result, pandas 3 returns an empty `DataFrame` rather than a `Series` — same family of quirk that #110-adjacent work just fixed for `cluster_phase`. Captured in the backlog comment on #116; not in scope here.

Closes #88. Part of #116.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Vg49h6F8yUBhT7yJEbn1pb)_

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,5 +7,6 @@ _version.py
 build/
 docs/_build/
 tests/integration/_plots/
+result_images/
 .ipynb_checkpoints
 *.pyc
--- a/tests/unit/plot/test_polygons.py
+++ b/tests/unit/plot/test_polygons.py
@@ -14,11 +14,11 @@ import numpy as np
 import pandas as pd
 import pytest
 from matplotlib.patches import Polygon
+from matplotlib.testing.decorators import check_figures_equal, remove_ticks_and_titles
 
 from landau import plot as plot_mod
 from landau.plot import get_phase_colors, get_polygons, plot_phase_diagram, plot_polygons
 from landau.poly import AbstractPolyMethod, Concave
-from matplotlib.testing.decorators import check_figures_equal
 
 
 # --- shared fixtures ---------------------------------------------------------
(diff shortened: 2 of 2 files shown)


pmrv:
@claude see if you can reuse some helpers from here to expand coverage https://matplotlib.org/stable/api/testing_api.html

pmrv:
Good idea, thanks — added two tests using `matplotlib.testing.decorators.check_figures_equal` in 9fb0ca6:

- `test_plot_polygons_renders_independent_of_input_order` — feeds the same polygons in opposite index orders into two figures and asserts the rendered images are identical. This expresses the zorder-by-bbox-area property as a visual invariant rather than an attribute check.
- `test_plot_phase_diagram_matches_explicit_pipeline` — renders `plot_phase_diagram(df, ax=ax)` against the explicit `get_polygons -> plot_polygons -> axis setup` decomposition. Documents what the public helper does in terms of the smaller helpers and guards against the two paths drifting apart in future refactors (e.g. the axis-setup extraction in #117).

I stuck to `check_figures_equal` rather than `image_comparison` to avoid checking baseline PNGs into the repo. Image extensions limited to `png` so we don't pull in the PDF/SVG backends in CI.

---

PR #127, opened 2026-05-13
Trim symmetric overlaps between buffered phase polygons

### LINKED ISSUE #125: Trim symmetric overlaps between buffered phase polygons
## Context

In `landau/poly.py` (`AbstractPolyMethod.make`, line ~85), every phase polygon is buffered by `min_c_width/2` so phases with very small solubility remain visible on the phase diagram. This is a feature worth keeping — without it some phases would be invisible or barely visible.

However, the buffer causes adjacent phases to overlap. The result looks messy and not crisp.

## Question I asked

Does shapely have a built-in feature to buffer all polygons and then symmetrically trim any overlap between them?

## Findings

There is no single shapely call for this, but it can be composed easily. Two reasonable idioms:

**1. Pairwise half-overlap subtraction (simple, ~5 lines)**

For each buffered polygon `a`, subtract `b.buffer(-eps/2)` (or, equivalently, half the intersection on the `b`-side) for every neighbour `b`. When the two originals are roughly equidistant across the seam — which is the common case for adjacent phase fields — the trimmed boundary lands at the midpoint. Fast and almost always good enough.

**2. Voronoi-based clipping (mathematically clean)**

Compute the generalized Voronoi diagram of the *original* (unbuffered) shapes — every point is assigned to its nearest original polygon — and clip each buffered polygon to its Voronoi cell. Shapely only ships point-based `shapely.ops.voronoi_diagram`, so the polygon version requires densifying boundaries into points, building the Voronoi diagram, `unary_union`-ing cells per phase, and intersecting with the buffered polygon. ~20 lines, but produces truly symmetric seams everywhere.

## Plan

Go with option 1. It's the lowest-risk change and visually indistinguishable from option 2 in the typical two-neighbour seam case in phase diagrams. If corner regions ever look asymmetric we can switch to option 2 later.

Implementation lives in `landau/poly.py`, applied in `AbstractPolyMethod.apply()` after all per-phase polygons are produced and before they are returned for plotting.

pmrv:
## Summary

- Keep the existing `min_c_width/2` buffer so small-solubility phases remain visible.
- After buffering, subtract each neighbour's un-buffered shape (`b.buffer(-r)`) from every polygon in `AbstractPolyMethod.apply`, so the visible seam lands on the originals' shared boundary instead of in the overlap strip.
- See #125 for the design rationale and a demonstrator script comparing this approach against generalized-Voronoi clipping.

## Test plan

- [x] `pytest tests/unit/test_poly.py` passes (the pre-existing hypothesis failure in `test_concave` on degenerate input is unrelated).
- [x] Visually inspect example phase diagrams from `docs/` notebooks; seams at adjacent phase boundaries should now be crisp.

Closes #125

---
_Generated by [Claude Code](https://claude.ai/code/session_01TCC9SncJBrNX49gQGu8HDy)_

### DIFF
--- a/landau/poly.py
+++ b/landau/poly.py
@@ -27,9 +27,10 @@ class AbstractPolyMethod(abc.ABC):
         over groups of columns `phase` and `phase_unit`."""
         return df
 
-    def make(self, dd: pd.DataFrame, variables: list[str] = ["c", "T"]) -> Polygon | None:
+    def make(self, dd: pd.DataFrame, variables: list[str] = ["c", "T"]) -> shapely.Polygon | None:
         """Turn the subset of the full data belonging to one phase region into
-        a polygon."""
+        a buffered shapely polygon.  Conversion to matplotlib happens in
+        :meth:`apply`."""
         border = dd["border"].to_numpy() if "border" in dd.columns else np.zeros(len(dd), dtype=bool)
         segment_label = dd["border_segment"].to_numpy() if "border_segment" in dd.columns else np.ones(len(dd), dtype=int)
 
@@ -86,7 +87,7 @@ class AbstractPolyMethod(abc.ABC):
         if isinstance(shape, shapely.MultiPolygon):
             shape = max(shape.geoms, key=shapely.area)
             warn("polymethod returned disjoined polygons, returning largest.")
-        return Polygon(shape.exterior.coords)
+        return shape
 
     @abc.abstractmethod
     def _make(self, pp: np.ndarray, border: np.ndarray, segment_label: np.ndarray) -> shapely.Geometry | None:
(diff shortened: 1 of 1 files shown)


landau/poly.py
pmrv, line 1: You may move the conversion from shapely polygons to mpl ones' from make to apply if that clears up the code

pmrv: Done in f9e2697 — `make` now returns the buffered shapely `Polygon`, `apply` runs `_trim_overlaps` on shapely geometries and converts to `matplotlib.patches.Polygon` at the end via a small `_to_mpl_polygon` helper. Removed the mpl→shapely→mpl round-trip in the trimmer; the code is noticeably shorter.

---

landau/poly.py
```diff
@@ -95,10 +96,55 @@ def _make(self, pp: np.ndarray, border: np.ndarray, segment_label: np.ndarray) -
...
+        shared boundary.
+        """
+        if len(shapes) < 2:
+            return shapes
+        r = self.min_c_width / 2
+        out: dict = {}
+        for k, a in shapes.items():
+            if a.is_empty:
+                out[k] = a
+                continue
```
pmrv: lets trim empty polygons early in apply

pmrv: Done in ff355cf — empties filtered right after `make().dropna()` in `apply`, so `_trim_overlaps` drops its inner `is_empty` checks.

---

pmrv:
rebase on main

PR #129, opened 2026-05-13
feat(poly): boundary-Voronoi cut for symmetric overlap trimming

pmrv:
## Summary

Implements [the boundary-Voronoi algorithm from the last comment on #125](https://github.com/eisenforschung/landau/issues/125#issuecomment-4443760642) in `AbstractPolyMethod._trim_overlaps`, replacing the previous plain "subtract every neighbour's un-buffered shape" pass that landed in #127.

For each overlapping pair `(a, b)`, per connected component of the buffered union:

1. Sample the union exterior uniformly into `B_U`.
2. Label each `B_U` point by which un-buffered original (recovered as `buffered.buffer(-r)`) is closer.
3. **Binary-search** each label transition along the boundary segment so the seam endpoint sits exactly where `d(oa) == d(ob)`.
4. **Route the cut through any shared geometry** of the two un-buffered originals (`oa.intersection(ob)` — a tangent corner, a coincident edge). The eutectic apex is at distance 0 from both, so the seam must pass through it. For point-tangent corners that erosion can't quite recover (the rounded buffer cap erodes back slightly off the original sharp tip), fall back to the midpoint of the closest-approach pair when it lies within `r`.
5. Split the union with the resulting polyline, hand each piece to its closer original, subtract the foreign piece from the other buffered polygon.

Falls back to the previous plain-difference behaviour where the bisector cut can't be constructed (one phase fully covers the component, the split is geometrically degenerate, or a phase is too thin to erode).

## Why

The simple subtract-the-neighbour's-original approach leaves a visible artefact at point-tangent corners (eutectic apices): the buffer rounds the sharp tip into a curve that pokes past the apex into the gap, creating a "tooth" on one side after subtraction. The boundary-Voronoi cut puts the seam at the true equidistance locus, which passes through the apex point.

Visual + timing comparison and full algorithm derivation are in [issue #125](https://github.com/eisenforschung/landau/issues/125) — see particularly the apex-zoom and timings table in [this comment](https://github.com/eisenforschung/landau/issues/125#issuecomment-4443903163). The reference demo lives on the `claude/voronoi-demo-issue-125` branch.

## Test plan

- [x] `tests/unit/test_poly.py::test_trim_overlaps_point_tangent_apex` — eutectic apex: trimmed shapes are symmetric, the apex is on the seam, residual overlap area ~0
- [x] `tests/unit/test_poly.py::test_trim_overlaps_shared_edge` — three rectangles tiling T-mu: every pairwise overlap is ~0 and no phase area is lost
- [x] `pytest tests/unit/test_poly.py tests/unit/plot/test_polygons.py` passes (the pre-existing `test_concave` failure on `shapely.concave_hull` for a pathological hypothesis input is independent of this change)
- [x] Manual visual check on the c-T eutectic scene shows the symmetric V-seam through `(0.5, 800)`

https://claude.ai/code/session_01WHKb72aizv9o6sep7jij7u

---
_Generated by [Claude Code](https://claude.ai/code/session_01WHKb72aizv9o6sep7jij7u)_

### DIFF
--- a/landau/poly.py
+++ b/landau/poly.py
@@ -153,9 +153,11 @@ class AbstractPolyMethod(abc.ABC):
                     continue
                 out[ki], out[kj] = _voronoi_split(a, b, oa, ob, inter, r)
 
+        # difference / split can return MultiPolygon (overlap split the
+        # shape in two) or a GeometryCollection mixing tiny LineString
+        # slivers with the main Polygon; keep the largest Polygon.
         for k, v in out.items():
-            if isinstance(v, shapely.MultiPolygon):
-                out[k] = max(v.geoms, key=shapely.area)
+            out[k] = _largest_polygon(v)
         return pd.Series(out, name=shapes.name).reindex(shapes.index)
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


pmrv:
check out 2d_toy_mu_phase_diagram, it doesn't seem to work reliably yet.

pmrv:
May work now, but not confident that this will extend to all diagrams.  I'm parking this here, in case we're not happy with option 1 in the future and want to reconsider.

PR #130, opened 2026-05-13
fix(poly): clean up apex residual after overlap subtraction

pmrv:
## Summary

A small cleanup pass on top of the symmetric subtract that landed in #127, addressing the only case it can't reach: two phases that touch at a single point (a eutectic apex). The rounded buffer cap extends past the apex into the gap and falls outside any neighbour's un-buffered shape, so the plain ``b.buffer(-r)`` subtraction leaves a small lens-shaped residual at the tip.

After the existing subtract pass, walk every still-overlapping pair and split each residual through the apex with a long line perpendicular to the segment between the two un-buffered originals' centroids (which bisects the lens symmetrically for the common case of two convex phases meeting head-on). Each half is removed from the polygon it doesn't belong to.

The cleanup is a no-op on:
- shared-edge cases (the subtract pass already lands the seam on the original boundary)
- residuals that are numerical noise — lines, points, or mixed collections from a buffer round-trip, which `shapely.ops.split` chokes on anyway

Also reduce every output to its largest Polygon component before returning: `Polygon.difference` / `shapely.ops.split` can return a `GeometryCollection` mixing the main polygon with tiny line/point slivers and the downstream `_to_mpl_polygon` only takes a single `Polygon` (without this, the toy T-μ testplot drops the s3 phase entirely).

This is the lightweight follow-up to the conversation on #125 / #129; the heavier boundary-Voronoi alternative is parked on `claude/voronoi-demo-issue-125` for reference.

## Test plan

- [x] `tests/unit/test_poly.py::test_trim_overlaps_apex_residual_split` — eutectic apex: residual overlap area ~0, symmetric trimmed areas, apex on the seam.
- [x] `pytest tests/unit/test_poly.py tests/unit/plot/test_polygons.py` passes (the pre-existing `test_concave` failure on a hypothesis edge case for `shapely.concave_hull` is unrelated).
- [x] Manual render of `tests/integration/testplots.py` shows `2d_basics`, `2d_basics_mu`, `2d_toy`, `2d_toy_mu` visually identical to main except the apex sliver is gone.

https://claude.ai/code/session_01WHKb72aizv9o6sep7jij7u

---
_Generated by [Claude Code](https://claude.ai/code/session_01WHKb72aizv9o6sep7jij7u)_

### DIFF
--- a/landau/poly.py
+++ b/landau/poly.py
@@ -7,6 +7,7 @@ from warnings import warn
 from pyiron_snippets.import_alarm import ImportAlarm
 
 import shapely
+import shapely.ops
 import numpy as np
 import pandas as pd
 from matplotlib.patches import Polygon
@@ -112,6 +113,11 @@ class AbstractPolyMethod(abc.ABC):
         creates an overlap strip; here we subtract from every polygon its
         neighbours' un-buffered shapes so the seam lands on the original
         shared boundary.
+
+        Where two phases touch at a single point (e.g. a eutectic apex)
+        the rounded buffer cap extends past the apex into the gap and
+        leaves a small residual overlap that the plain subtract can't
+        reach; :func:`_split_apex_residual` finishes those off.
         """
         if len(shapes) < 2:
             return shapes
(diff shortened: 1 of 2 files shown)


pmrv:
gonna park it too until it actually becomes annoying.

PR #140, opened 2026-05-19
fix: defer AsePhase construction via fixture in test_vectorize

### LINKED ISSUE #133: Defer ASE `AsePhase` construction in `test_vectorize.py` parametrize
`tests/unit/phases/test_vectorize.py:13-19` calls `get_ase_phase()` at parametrize time, and `get_ase_phase` references `AsePhase` whose import is conditional on the `ImportAlarm` block at the top of the file:

```python
with ImportAlarm() as ase_alarm:
    from ase.thermochemistry import HarmonicThermo
    from landau.phases.asewrapper import AsePhase

def get_ase_phase():
    if ase_alarm.message is None:
        return AsePhase(...)   # <- NameError if ASE is absent
    return None

@pytest.mark.skipif(ase_alarm.message is not None, reason="ASE is not installed")
@pytest.mark.parametrize("phase", [get_ase_phase()])
```

`pytest.mark.parametrize` evaluates its `argvalues` at collection time, **before** `skipif` is consulted, so on a system without ASE the module fails to *collect* (NameError) rather than skipping cleanly. The module-wide skip marker that #111 added (referenced in the working-style notes) does not help here for the same reason.

## Proposed fix

Defer the `AsePhase` construction. Two natural shapes:

1. Wrap the body with `pytest.param(..., marks=pytest.mark.skipif(...))` and provide a lazy callable / id, or
2. Convert `phase` to a fixture that builds `AsePhase` inside the function body (where `skipif` has already taken effect).

Option (2) is the simpler, more idiomatic pytest pattern:

```python
@pytest.fixture
def ase_phase():
    return AsePhase("ht_phase", 0.5, thermochem=HarmonicThermo(vib_energies=[0.1]), pressure=None)

@pytest.mark.skipif(ase_alarm.message is not None, reason="ASE is not installed")
@given(...)
def test_semigrand_potential_vectorization(ase_phase, data):
    ...
```

## Tests / acceptance

- `pytest tests/unit/phases/test_vectorize.py --co` collects cleanly on a machine without ASE installed (skips rather than errors).
- The two tests still run identically when ASE *is* installed.

## Scope

- Single file change.

Part of #116.

pmrv:
Replace `get_ase_phase()` + `@pytest.mark.parametrize` with an `ase_phase` fixture so that `AsePhase(...)` is never evaluated at collection time when ASE is absent.

Closes #133.

### DIFF
--- a/tests/unit/phases/test_vectorize.py
+++ b/tests/unit/phases/test_vectorize.py
@@ -1,7 +1,7 @@
 import pytest
 import numpy as np
 
-from hypothesis import given, strategies as st
+from hypothesis import HealthCheck, given, settings, strategies as st
 from hypothesis.extra.numpy import arrays, mutually_broadcastable_shapes
 
 from pyiron_snippets.import_alarm import ImportAlarm
@@ -18,6 +18,7 @@ def ase_phase():
 
 @pytest.mark.skipif(ase_alarm.message is not None, reason="ASE is not installed")
 @given(data=st.data())
+@settings(suppress_health_check=[HealthCheck.function_scoped_fixture])
 def test_semigrand_potential_vectorization(ase_phase, data):
     shapes = data.draw(mutually_broadcastable_shapes(num_shapes=2, max_dims=3, max_side=3))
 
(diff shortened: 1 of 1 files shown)


pmrv:
@claude babysit hypothesis tests.

PR #145, opened 2026-05-21
feat: add plot_excess_free_energy and ExcessFreeEnergy notebook

### LINKED ISSUE #136: Plot for 'Common Tangent Like'
excess free energy vs concentration, potentially at a few temperature.

prabhath-c:
Add plot_excess_free_energy() to landau/plot.py and export from landau/__init__.py. Helpers: _phase_concentration_range, _endmember_free_energy (auto-detects end members from phases covering c=0/c=1), _lower_convex_hull (Andrew's monotone chain for convex hull).

Line phases render as scatter points; solution phases as curves; dashed black line shows the lower convex hull (common-tangent construction).

Notebook demonstrates 3 SlowInterpolatingPhase curves + 1 LinePhase point. Executed with embedded outputs.

Closes #136


prabhath-c:
@claude 

* In the branch claude/issue-136-20260521-1020, make the plotting of the convex hull optional and when true, along with the convex hull, mark the points making the hull with black scatter points. Implement this. 
* Additionally, what kind of unit tests can be written for this function? Discuss with me before implementing the tests.

prabhath-c:
@claude Rerun the notebook with the updated function and use only the case of `convex_hull = True` for the notebook.

On a separate pull request that merges into this very branch and hence this pull request, write the tests you think you should. Call the branch tests/excess_free_energy_plots

prabhath-c:
@claude I found a much better way to plot it, by looking at the intermetallics.ipynb.

```python
import pandas as pd

df = pd.concat([
    calc_phase_diagram([solid, liquid, inter], Ts=T, mu=100, keep_unstable=True)
        for T in [500, 1000, 1600]
], ignore_index=True)

g = sns.relplot(
    data=df,
    x='c', y='f_excess',
    hue='phase',
    col='T',
    style='stable', style_order=[True, False]
).map_dataframe(
    lambda data, *a, **kw: sns.lineplot(data.query('border'), *a, **kw),
    x='c', y='f_excess',
    ls='dotted', 
    units='mu', estimator=None,
    color='k',
).map_dataframe(
    lambda data, *a, **kw: sns.scatterplot(data.query('border'), *a, **kw),
    x='c', y='f_excess',
    color='k',
    zorder=2,
).refline(
    y=0
).set(
    xlabel='Concentration',
    ylabel='Free Energy of Formation',
)
```

The graphs also look much better. Implement this into the function and remove the current way of doing it. Keep the important input arguments like convex_hull = True and color override options. If convex hull = False, then instead of scatter, make it a curve plot (apart from the line phases, which will reflect as a dot), and if True, then the stable parts will be a solid curve, while the unstable ones will be a dotted line. The input arguments will still be the list of phases and temps, and not the df. The df will be internally constructed.

Update and run the notebook as well. Use a similar example to the one in the intermetallics.ipynb. It shows clearly the competition of phases. Add another intepolating phase too, so see how well the function can handle complexity. Run the funciton for 3 temperatures only, so that the plots are big enough. Also, I think the title of the full plot was interfereing with the titles of the subplots

prabhath-c:
@claude, I asked you to use the example used in the:

```python

solid_a = LinePhase('A', fixed_concentration=0, line_energy=-2.0, line_entropy=1.0*kB)
solid_b = LinePhase('B', fixed_concentration=1, line_energy=-3.0, line_entropy=1.5*kB)

solid = IdealSolution('solid', solid_a, solid_b)

liquid_a = LinePhase('A(l)', fixed_concentration=0, line_energy=-1.9, line_entropy=2.5*kB)
liquid_b = LinePhase('B(l)', fixed_concentration=1, line_energy=-2.9, line_entropy=2.2*kB)

liquid = IdealSolution('liquid', liquid_a, liquid_b)

inter = LinePhase('AB$_2$', fixed_concentration=2/3, line_energy=-2.8, line_entropy=1.3*kB)
```

This example is there in the https://github.com/eisenforschung/landau/blob/main/notebooks/Intermetallics.ipynb

prabhath-c:
@claude The example has been corrected, but the convex hull is not plotting correctly as show in the attached image. It is not only plotting it as a colored line, but it is also missing the intermetallic point. Also, when the phase is not stable, mark it by scatter points, not a dashed line. Black dotted line is only for the convex hull. Be very careful and go through the function and the other notebook to make sure the output is clear.

grafik

prabhath-c:
grafik

@pmrv, the end points of the stable segments are shorting and forming a line by themselves. Do you happen to recall how we solved this for the plotting functions earlier?

prabhath-c:
@claude Look at the plots attached and the plotting functions we have. The plots show that there are stable points across two separate segments that seem to be shorting and forming a straight line.

Also, use a darker color palette like the `sns.color_palette("pastel")` and make the points slightly darker and the curves slightly thicker. Also, the convex hull vertex for the intermetallic point isn’t visible - possible that the actual point is on top of the convex vetex.

grafik

pmrv:
> grafik
> 
> @pmrv, the end points of the stable segments are shorting and forming a line by themselves. Do you happen to recall how we solved this for the plotting functions earlier?

In the Intermetallics notebook I avoided it by just scatter plotting all points.  In the 1D plots we added labels for every segment and then used `units` keyword to seaborn.  We can try to get the labels with [one of the cluster functions](https://github.com/eisenforschung/landau/blob/0e1790e122fbf981328d280497e4cc85440d9c71/landau/calculate.py#L263).

pmrv:
> > grafik
> > @pmrv, the end points of the stable segments are shorting and forming a line by themselves. Do you happen to recall how we solved this for the plotting functions earlier?
> 
> In the Intermetallics notebook I avoided it by just scatter plotting all points. In the 1D plots we added labels for every segment and then used `units` keyword to seaborn. We can try to get the labels with [one of the cluster functions](https://github.com/eisenforschung/landau/blob/0e1790e122fbf981328d280497e4cc85440d9c71/landau/calculate.py#L263).

@claude try and use the cluster method I mention to generate segment ids.

prabhath-c:
@claude Much better. Now, increase the thickness of the curves and the points on the phases. The intermetallic point is slightly too large. I want the points to be darker and more visible. For the vertices of the convex hull, use a slightly smaller point and a filled circle, not a transparent one. The vertex point has to be smaller than the line phase point itself.

prabhath-c:
@pmrv The cluster method seems to work to a certain extent. The 1600K plot still has those segments. Also the convex hull is connecting all the points, which is different from what we have in the last cell of https://github.com/eisenforschung/landau/blob/main/notebooks/Intermetallics.ipynb. Do we want it to be fully connected?

grafik

PS: This isn't urgent. Reply when you are free.

pmrv:
@claude debug why the clustered segment labels are still lumping together both stable branches.  you may need to tune cluster thresholds or use that other cluster function in T, c, mu space we have somewhere.  Start by writing a unit test that applies the cluster function on the example data you have in the notebook.  Work in a new PR that targets this one.

pmrv:
@claude based on your diagnosis, create a new PR off main that exposes the cluster threshold as a function parameter.

prabhath-c:
> ### Tasks
> 
>     * [x]  Gather context
> 
>     * [x]  Read `calculate.py` cluster functions
> 
>     * [x]  Read `plot.py` current segmentation logic
> 
>     * [x]  Understand why clustering lumps stable branches together
> 
>     * [ ]  Write unit test using notebook example data (demonstrates the bug)
> 
>     * [ ]  Implement fix: per-(T, phase) gap-based segmentation
> 
>     * [ ]  Create new branch `fix/excess-free-energy-segment-clustering` targeting `claude/issue-136-20260521-1020`
> 
>     * [ ]  Commit, push, open PR
> 
> 
> ### Root cause
> 
> At T=1600K the stable liquid phase has **two disconnected concentration segments** (c∈[0, 0.501] and c∈[0.808, 0.955]) separated by a gap of ~0.307 where sigma is stable. `cluster_T_c` uses `distance_threshold=0.5` (hand-tuned for 2D T-c diagrams). In 1D concentration space at fixed T, any gap smaller than 0.5 is missed — that's the entire problem.
> 
> `cluster_T_c_mu` does not help either: the 3D distance between the two closest boundary points (c gap 0.307, mu gap 0.198) is √(0.307² + 0.198²) ≈ 0.365, still below 0.5.
> ### Fix
> 
> Replace the `cluster_phase(base_data)` call in `plot_excess_free_energy` with a local helper `_cluster_stable_segments` that works per-(T, phase) group: sort by c, split on gaps larger than `max(0.05, 5 × 90th-percentile spacing)`. This correctly identifies both liquid segments at T=1600K without touching any shared clustering infrastructure.
> 
> 
> [View job run](https://github.com/eisenforschung/landau/actions/runs/26247727988)

@claude Finish this job for this branch. I want to see in the notebook if the local helper changes anything

prabhath-c:
@pmrv Finally! It works! Did not expect the function to get this long just because I wanted to make the convex hull optional.
Are you satisfied with the function? I suppose more functionality can be added as and when required.

grafik

pmrv:
> @pmrv Finally! It works! Did not expect the function to get this long just because I wanted to make the convex hull optional. Are you satisfied with the function? I suppose more functionality can be added as and when required.
> grafik

I'll do some clean up on it tomorrow.  I'll definitely want it to use the existing cluster methods contingent on #153 and I'm not sure I like that it combines calculating and plotting.  Right now I feel it'd be a lot simpler to have callers pass the pre computed 'phase diagram'.

prabhath-c:
> 
> I'll do some clean up on it tomorrow. I'll definitely want it to use the existing cluster methods contingent on #153 and I'm not sure I like that it combines calculating and plotting. Right now I feel it'd be a lot simpler to have callers pass the pre computed 'phase diagram'.

I was initially thinking of it taking a calculated `df` as an input too. But calculating the `df` itself takes time with the `SlowInterpolatingPhase`. This function could then serve as a possible diagnostic before beginning the full calculation. Upto you, I have no strong feelings about it :)

pmrv:
@claude rebase on main and try again to use the existing cluster function

pmrv:
> > I'll do some clean up on it tomorrow. I'll definitely want it to use the existing cluster methods contingent on #153 and I'm not sure I like that it combines calculating and plotting. Right now I feel it'd be a lot simpler to have callers pass the pre computed 'phase diagram'.
> 
> I was initially thinking of it taking a calculated `df` as an input too. But calculating the `df` itself takes time with the `SlowInterpolatingPhase`. This function could then serve as a possible diagnostic before beginning the full calculation. Upto you, I have no strong feelings about it :)

But that's exactly why I want it outside, so the calculation is only done once and plots can be done without the expensive phase calls.

pmrv:
@claude set `'units='mu'` for the dotted plots of the stable tangents.  I want lines for the metastable branches of the free energy.  same color as the stable ones but with some alpha applied.  rerun notebook before commiting

pmrv:
@claude the black border markers are overshadowing markers for intermetallics of the same size.  former go smaller (but larger zorder) and latter go bigger

pmrv:
@claude split calculation and plotting,  keep calculation in notebook.

pmrv:
@claude there's a bug in the calculation of the excess free energy visible in the left most plot.  The lowest point at c -> 0 can only be 0.   Likely the green phase is taken as the reference at the wrong temperature and kept for all.  Is this a bug in calc_phase_diagram or the new plotting?

pmrv:
@claude document in a new issue then

pmrv:
@claude add this plot to the testplot script then squash all commits.

pmrv:
@prabhath-c lgtm unless you have more plans for it.

pmrv:
@claude go over existing notebooks to apply the new function

pmrv:
@claude ensure new method also works for single temperatures.

prabhath-c:
@pmrv Thanks for fixing this function!

How do you extract the images from the tests to get the plots here in the pull request itself?

pmrv:
@prabhath-c I had claude write [this action](https://github.com/eisenforschung/landau/blob/main/.github/workflows/testplot.yml)  There's some scripts [here](https://github.com/eisenforschung/landau/tree/main/.github/scripts) that duplicate the plotting code from the tests.  The action then commits the images to a burner [branch](https://github.com/eisenforschung/landau/tree/testplots) and links them in the response.

PR #149, opened 2026-05-21
Sandbox the @testplot mention workflow and document the trade-off

pmrv:
Splits the `@testplot` mention workflow into three jobs so PR-controlled code never runs in the same context as repo write secrets, and updates the docstring in `tests/integration/testplots.py` to record the resulting trade-off.

## Workflow split

- `parse` — checks out `main`, installs the Claude Code CLI, runs `.github/scripts/parse_testplot_mention.py` with `CLAUDE_CODE_OAUTH_TOKEN` and the comment body. Emits `sha`, `args_json`, and `note` as job outputs.
- `render` — checks out the PR head with `persist-credentials: false`, declares `permissions: contents: read`, and runs `pip install -e .[fast-tsp,python-tsp]` + `python tests/integration/testplots.py` against the PR's code. No secrets are referenced in any step env. Uploads `_plots/` as an artifact.
- `publish` — downloads the artifact, clones the repo with `GITHUB_TOKEN`, pushes the PNGs to the `testplots` gallery branch, and posts the inline-image comment. Runs no PR code.

## Trade-off (documented in `testplots.py`)

Parser allow-list of plot keys is frozen to `main`. A PR that adds a new plot to `testplots.py` can't be exercised via `@testplot` until a parser-only PR lands the new key on `main` first.

## Test plan

- [x] yaml parses; `jq -c` round-trip of `args_json` works for empty and populated arg lists
- [x] `permissions: contents: read` on the `render` job; no `secrets.*` references in any of its step envs
- [ ] Land and exercise `@testplot 2d fasttsp` on a follow-up PR; confirm the artifact handoff between `render` and `publish` and the published comment

### DIFF
--- a/.github/workflows/testplot-mention.yml
+++ b/.github/workflows/testplot-mention.yml
@@ -8,6 +8,21 @@ name: testplot-mention
 # Plots are published to the same content-addressed `testplots` gallery
 # branch used by .github/workflows/testplot.yml so a single PR can
 # accumulate diagrams from both the label-driven and mention-driven flows.
+#
+# Job split to keep the PR's untrusted code away from repo write secrets:
+#   parse    — checks out `main` (trusted), runs the Haiku parser with
+#              CLAUDE_CODE_OAUTH_TOKEN, emits the parsed args as outputs.
+#   render   — checks out the PR head (untrusted), installs and runs the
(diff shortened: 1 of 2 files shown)


pmrv:
@testplot only the prettiest of plots

PR #156, opened 2026-05-22
fix(calculate): seed guess_mu_range from coarse scan instead of mu=0

### LINKED ISSUE #31: automatic mu range brittle at low temperatures

pmrv:
Closes #31

## Problem

`guess_mu_range` used BFGS with `x0=[0]` to find the mu values that minimise and maximise `c(mu)`. At low T, `c(mu)` is a step function: the gradient is ~0 almost everywhere and undefined at the jump. BFGS returns immediately from `x0=0` without moving, so `mu0 == mu1`, triggering the tolerance-reduction recursion without ever converging. Even when the transition is near `mu=0`, the step-function `cc` returned by `c(linspace(mu0, mu1))` has repeated values that `interp1d` cannot handle (requires strictly monotone x).

## Fix

- Scan `linspace(-10, 10, 200)` to locate the seeds for the BFGS minimise/maximise calls. These seeds are close to the true extrema regardless of temperature.
- Sort and deduplicate `cc` before `interp1d` to handle the step-function case at very low T.

## Tests added (`tests/unit/test_calculate.py`)

- `test_guess_mu_range_returns_samples` — output length matches `samples` argument
- `test_guess_mu_range_covers_concentration_space` — `c0 < 0.1`, `c1 > 0.9` at normal T
- `test_guess_mu_range_low_temperature` — succeeds at T=10 K where the old code hung/raised
- `test_guess_mu_range_mu_not_centred_at_zero` — transition shifted to mu≈5 eV, old `x0=0` seed would be in the flat region
- `test_guess_mu_range_degenerate_raises` — same-concentration phases still raise `ValueError`

198 passed, 5 skipped (unchanged from main).

https://claude.ai/code/session_01NHe6e2HGLVzBiGfV3emryr

---
_Generated by [Claude Code](https://claude.ai/code/session_01NHe6e2HGLVzBiGfV3emryr)_

### DIFF
--- a/landau/calculate.py
+++ b/landau/calculate.py
@@ -134,18 +134,8 @@ def guess_mu_range(phases: Iterable[Phase], T: float, samples: int, tolerance: f
         ci = (prob * conc).sum(axis=0)
         return ci
 
-    # Coarse scan over a wide mu range to find seeds for BFGS.  At low T,
-    # c(mu) is nearly a step function with zero gradient almost everywhere,
-    # so starting BFGS from mu=0 may not move at all.
-    mu_scan = np.linspace(-10, 10, 200)
-    cc_scan = c(mu_scan)
-    mu0_seed = mu_scan[np.argmin(cc_scan)]
-    mu1_seed = mu_scan[np.argmax(cc_scan)]
-
-    resi = so.minimize(lambda x: +c(x[0]), x0=[mu0_seed], tol=tolerance, method="BFGS")
-    resa = so.minimize(lambda x: -c(x[0]), x0=[mu1_seed], tol=tolerance, method="BFGS")
-    mu0 = resi.x[0]
-    mu1 = resa.x[0]
+    mu0 = so.brute(lambda x: +c(x[0]), ranges=[(-10, 10)], Ns=200)[0]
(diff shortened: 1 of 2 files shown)


landau/calculate.py
```diff
@@ -134,8 +134,16 @@ def c(mu):
...
+    # Coarse scan over a wide mu range to find seeds for BFGS.  At low T,
+    # c(mu) is nearly a step function with zero gradient almost everywhere,
+    # so starting BFGS from mu=0 may not move at all.
+    mu_scan = np.linspace(-10, 10, 200)
+    cc_scan = c(mu_scan)
+    mu0_seed = mu_scan[np.argmin(cc_scan)]
+    mu1_seed = mu_scan[np.argmax(cc_scan)]
+
+    resi = so.minimize(lambda x: +c(x[0]), x0=[mu0_seed], tol=tolerance, method="BFGS")
+    resa = so.minimize(lambda x: -c(x[0]), x0=[mu1_seed], tol=tolerance, method="BFGS")
```
pmrv: use the bruteforce global optimizer from scipy

pmrv: @claude

PR #163, opened 2026-05-26
feat(refine): add boundary_id column to refined rows

### LINKED ISSUE #124: Tag refined rows with an explicit phase-boundary id
**Background**

The Clausius-Clapeyron refiners (`ClausiusClapeyronRefiner`,
`MiscibilityGapRefiner`) produce one refined transition per converged
isothermal step, and the row(s) emitted into the output dataframe
share the exact `(T, mu)` of that transition. At present the only way
a consumer (e.g. plotting code) can tell which rows belong to the
same coexistence line is by grouping on `(T, mu)` and walking the
result by hand — a workaround that re-discovers information the
refiner had on hand all along.

**Request**

Add an explicit "line id" or "boundary id" tag to refined rows so
they can be grouped directly:

* `RefinedPoint` / `RefinedMiscibilityGap` carry a stable identifier
  for the coexistence line they belong to.
* `_CCBase.run` (and any other line-tracing refiners we add later)
  assigns one id per traced line and propagates it to `to_rows()`.
* The id ends up as a column on the output dataframe, e.g.
  `boundary_id`.

This subsumes a related debug-only mechanism that was used during
development of #115 (a `seed` column tagging which propose-time
candidate produced each row). That `seed` mechanism has been
removed; the cleaner replacement here is what we actually want.

**Acceptance**

* Notebook plotting code groups by the new column instead of
  inferring lines from `(T, mu)` clusters.
* No behavioural change to the refined `(T, mu, phi, c)` values.

Context: pmrv's review on PR #115
(<https://github.com/eisenforschung/landau/pull/115#discussion_r3231118162>).

pmrv:
Closes #124.

## What

`RefinedPoint` and `RefinedMiscibilityGap` gain a `boundary_id: int = 0` field. `to_rows()` propagates it into every emitted dict row, so the output DataFrame carries a `boundary_id` column that callers can `groupby` directly instead of inferring coexistence lines from `(T, mu)` clusters.

## Assignment logic

- **`_CCBase.run()`** — one id per `_pair_key`: all rows from the same traced coexistence line (same phase pair for `ClausiusClapeyronRefiner`, same phase name for `MiscibilityGapRefiner`) share one `boundary_id`, even when multiple simplices seed the same line (the straddle dedup keeps them on the same key).
- **Base `Refiner.run()`** — one id per non-empty `solve()` call. For point-finding refiners (`ScanRefiner`, `DelaunayLineRefiner`, `DelaunayTripleRefiner`) each candidate result gets its own id.

## Backward compatibility

`boundary_id` defaults to `0`, so existing call sites that construct `RefinedPoint(T=..., mu=..., phases=...)` without it still work.

## Tests

Six new tests in `tests/unit/test_refine.py`:

- `test_boundary_id_refined_point_to_rows` — `to_rows` propagates `boundary_id` into rows
- `test_boundary_id_refined_miscibility_gap_to_rows` — same for `RefinedMiscibilityGap`
- `test_boundary_id_cc_refiner_single_line` — single two-phase system → one unique id
- `test_boundary_id_cc_refiner_two_lines` — three-phase system → at least two distinct ids, each group has at most two phases
- `test_boundary_id_miscibility_gap_refiner` — regular solution gap → one unique id
- `test_boundary_id_delaunay_triple_rows_share_id` — triple point → all three rows share one id

All 209 unit and regression tests pass.

https://claude.ai/code/session_0187kmHM8LZdHRxgZYR23Hd3

---
_Generated by [Claude Code](https://claude.ai/code/session_0187kmHM8LZdHRxgZYR23Hd3)_


pmrv:
@claude add a test that compares this tag with the output of the existing cluster method.  there shouldn't be cases where one of the old cluster labels spans multiple boundary tags, though the inverse is ok.

pmrv:
@claude mh, useless test then.  Revert then fix merge conflicts.  Afaict the new tag is not used when constructing polygons.  That's out of scope for now. Report back in the original issue what would need to be done to implement that and some ideas for unit tests.

pmrv:
@claude there's still a merge conflict in claude.md

PR #166, opened 2026-05-28
fix: use most stable endpoint phase as f_excess reference

### LINKED ISSUE #155: bug: f_excess reference in calc_phase_diagram picks wrong endpoint phase
## Bug description

`calc_phase_diagram` computes `f_excess` via an inner `sub` function that selects the reference free energies at c=0 and c=1 by finding the global minimum concentration across all phases:

```python
c0 = dd.c.min()                    # smallest c in the whole dataset
f0 = dd.query("c==@c0").f.min()    # f of whichever phase has that c
```

This is wrong in two ways:

1. **Wrong phase selected as reference.** When two solution phases both sample near c=0, the one that happens to sample a slightly smaller `c` wins, regardless of which is thermodynamically more stable at the endpoint. At T=500K in the Intermetallics example, liquid samples c=2.26e-6 while solid samples c=5.04e-6, so liquid is used as the c=0 reference even though solid is more stable there (f_solid ≈ −2.043 eV vs f_liquid ≈ −2.008 eV). This shifts every `f_excess` value by ≈0.035 eV, producing negative `f_excess` for the solid phase (thermodynamically impossible).

2. **Line phases at non-endpoint concentrations can corrupt the reference.** A `LinePhase` at fixed concentration (e.g. AB₂ at c=2/3) still samples a finite `c` range due to numerical evaluation near its fixed point. If that minimum c happens to be smaller than the solution phases' minimum c, and the line phase has a large negative `f` (which it will at extreme μ values, since its f is linear in μ), it will be picked as `f0` and corrupt the entire `f_excess` calculation.

## Reproducer

```python
from landau import *
import numpy as np

kB = 8.617e-5

solid_a = LinePhase('A',   fixed_concentration=0, line_energy=-2.0, line_entropy=1.0*kB)
solid_b = LinePhase('B',   fixed_concentration=1, line_energy=-3.0, line_entropy=1.5*kB)
solid   = IdealSolution('solid', solid_a, solid_b)

liquid_a = LinePhase('A(l)', fixed_concentration=0, line_energy=-1.9, line_entropy=2.5*kB)
liquid_b = LinePhase('B(l)', fixed_concentration=1, line_energy=-2.9, line_entropy=2.2*kB)
liquid   = IdealSolution('liquid', liquid_a, liquid_b)

df = calc_phase_diagram([solid, liquid], Ts=500, mu=100, keep_unstable=True)

# c -> 0 limit: solid should be more stable (lower free energy)
# so f_excess should be 0 for solid and > 0 for liquid at c=0
# but current code may give negative f_excess for solid
c0_rows = df[df['c'] < 0.01].groupby('phase')['f_excess'].min()
print(c0_rows)
# Expected: solid ≈ 0, liquid > 0
# Actual:   solid < 0, liquid ≈ 0  (wrong)
```

## Expected behaviour

At the c=0 endpoint, the most stable phase should have `f_excess = 0` (it is the reference). All other phases should have `f_excess >= 0` there (they are metastable or unstable). Same applies at c=1.

## Proposed fix

Compute `f0` and `f1` per phase — find each phase's `f` at its own `c_min`/`c_max` — then take the minimum only over phases whose concentration range reaches the endpoint (within some tolerance, e.g. 10% of total c span):

```python
def sub(dd):
    dd = dd.query("-inf<mu<inf")
    c_lo, c_hi = dd.c.min(), dd.c.max()
    lo_thr = c_lo + 0.1 * (c_hi - c_lo)
    hi_thr = c_hi - 0.1 * (c_hi - c_lo)
    f0, f1 = np.inf, np.inf
    for _, g in dd.groupby("phase"):
        c_min_p, c_max_p = g["c"].min(), g["c"].max()
        if c_min_p <= lo_thr:
            f0 = min(f0, g.loc[g["c"].idxmin(), "f"])
        if c_max_p >= hi_thr:
            f1 = min(f1, g.loc[g["c"].idxmax(), "f"])
    if not np.isfinite(f0):
        f0 = dd.loc[dd["c"].idxmin(), "f"]
    if not np.isfinite(f1):
        f1 = dd.loc[dd["c"].idxmax(), "f"]
    return dd.f - (f0 * (1 - dd.c) + f1 * dd.c)
```

Identified in PR #145.

pmrv:
Closes #155.

## Problem

`calc_phase_diagram` computes `f_excess` via `sub()`, which selected the c=0 reference as:

```python
c0 = dd.c.min()                      # globally smallest c across all phases
f0 = dd.query("c==@c0").f.min()      # f of whichever phase has that c
```

When two solution phases both sample near c=0 (e.g. c=2.26e-6 vs c=5.04e-6 in the Intermetallics notebook), the one with the slightly smaller numerical value wins — regardless of thermodynamic stability. The metastable phase incorrectly becomes the reference, producing a negative `f_excess` for the stable phase at the endpoint.

## Fix

In `sub()`, iterate over phases. For each phase whose `c_min` lies within 1% of the full concentration span from the endpoint, record its `f` at its own minimum concentration. Take the minimum `f` across qualifying phases as `f0`/`f1`. Falls back to the global minimum if no phase qualifies.

```python
lo_thr = c_min + 0.01 * c_span
...
for _, g in dd.groupby("phase"):
    if g["c"].min() <= lo_thr:
        f0 = min(f0, g.loc[g["c"].idxmin(), "f"])
```

The threshold is 1%, not 10%: a 10% band would accept a line phase fixed at c=0.08 as a c=0 reference. At large negative μ such a phase has f = line_energy (constant), which can be much lower than the solution phase endpoint energy, corrupting f0 and shifting every f_excess value in the column.

## Tests

Four regression tests in `tests/regression/test_issue_155.py`:

- `test_stable_phase_at_c0_has_near_zero_f_excess` — solid f_excess at c<0.01 is > −0.1 eV (old bug: ≈ −0.5 eV; fixed: ≈ kT·mixing ≈ 0)
- `test_metastable_phase_at_c0_has_positive_f_excess` — liquid f_excess at c<0.01 is > 0 (metastable phase above the reference)
- `test_metastable_phase_f_excess_exceeds_stable_at_c0` — liquid f_excess exceeds solid f_excess near c=0 at every sampled point
- `test_line_phase_near_endpoint_not_used_as_reference` — `LinePhase` at c=0.08 with line_energy=−5 eV does not corrupt f0; solid f_excess stays ≈ 0 (with 10% threshold it would be ≈ +3 eV)

All 220 tests pass (5 skipped, unchanged).


tests/regression/test_issue_155.py
```diff
@@ -0,0 +1,137 @@
...
+When two solution phases both sample near c=0, the old code selected the phase
+with the globally smallest sampled concentration as the f_excess reference, even
+when a different phase has a lower free energy there.  The fix evaluates each
+phase at its own endpoint concentration and takes the thermodynamically stable one
+as the reference.
+
+The endpoint threshold is 1% of the full concentration span.  10% was too loose:
+a line phase fixed at c=0.08 (within 10% of c=0) could be accepted as a c=0
+reference when it happens to have a low free energy at extreme μ, corrupting the
+entire f_excess column.
```
pmrv: @claude remove the reference to 10% threshold since this never reflected in the current source.  Update your CLAUDE.md as well to make a note that comments and docstrings should only reflect the state of the code as it is, and maybe some historical justification, but *never* reference the convoluted windings within one PR.

tests/regression/test_issue_155.py
```diff
@@ -0,0 +1,128 @@
...
+
+    liquid_a = LinePhase("A(l)", fixed_concentration=0, line_energy=-1.5, line_entropy=0)
+    liquid_b = LinePhase("B(l)", fixed_concentration=1, line_energy=-2.5, line_entropy=0)
+    liquid = IdealSolution("liquid", liquid_a, liquid_b)
+
+    return solid, liquid
+
+
+@pytest.fixture
+def phase_diagram(two_phase_system):
```
pmrv, line 41: I forget, are the fixture module level?  If not, set them such to avoid recomputing them for every test.

tests/regression/test_issue_155.py
```diff
@@ -0,0 +1,128 @@
+"""Regression tests for #155.
+
+When two solution phases both sample near c=0, the old code selected the phase
+with the globally smallest sampled concentration as the f_excess reference, even
+when a different phase has a lower free energy there.  The fix evaluates each
```
pmrv: not clear enough , maybe
```suggestion
when a different phase has a lower free energy nearby.  The fix evaluates each
```

pmrv (commented):
@claude revise

PR #174, opened 2026-06-02
fix(poly): handle_poly_method TypeError when ratio= passed directly

### LINKED ISSUE #172: `handle_poly_method`: explicit `ratio=` keyword collides with the `alpha` back-compat alias
Refs #116.

## Context

`landau/poly.py:518-528`:

```python
def handle_poly_method(poly_method, **kwargs):
    ratio = kwargs.pop('alpha', Concave.ratio)
    allowed = {
                'concave': Concave(**kwargs, ratio=ratio),
                ...
    }
```

If a caller passes `ratio=` directly (legitimate, since `Concave` declares `ratio: float = 0.1`), `kwargs` retains it and the explicit `ratio=ratio` keyword causes a `TypeError: Concave.__init__() got multiple values for keyword argument 'ratio'`.

```python
>>> handle_poly_method('concave', ratio=0.3)
TypeError: ... multiple values for keyword argument 'ratio'
```

The `alpha` path was the legacy spelling kept alive by `@deprecate(alpha=...)` on `plot_phase_diagram` / `plot_mu_phase_diagram`. Both the deprecated `alpha` and the canonical `ratio` should work without colliding.

## Proposal

```python
ratio = kwargs.pop('alpha', kwargs.pop('ratio', Concave.ratio))
```

or equivalent — pop `ratio` first if present, then let `alpha` override it for the deprecated path. Update `Concave` and `Segments` construction accordingly.

## Test

Add two cases to `tests/unit/test_poly.py::test_handle_poly_method`:

- `handle_poly_method('concave', ratio=0.3).ratio == 0.3`
- `handle_poly_method('concave', alpha=0.4).ratio == 0.4` (already implicitly covered, make it explicit)

Then a regression case asserting no `TypeError`.

## Scope

Self-contained, no API change for any current caller (no one passes `ratio=` today, see `grep`), purely fixes the kwargs collision.

pmrv:
Closes #172.

## Problem

`handle_poly_method('concave', ratio=0.3)` raised `TypeError: Concave.__init__() got multiple values for keyword argument 'ratio'` because `ratio` was left in `**kwargs` while also being passed as the explicit `ratio=ratio` keyword.

## Fix

One-line change in `handle_poly_method` (`landau/poly.py`):

```python
# before
ratio = kwargs.pop('alpha', Concave.ratio)

# after
ratio = kwargs.pop('alpha', kwargs.pop('ratio', Concave.ratio))
```

`ratio` is now popped from `kwargs` before forwarding to `Concave()`. The deprecated `alpha=` spelling continues to work and takes precedence over `ratio=` when both are supplied (consistent with the existing deprecation intent).

## Tests

Two new assertions added to `test_handle_poly_method` in `tests/unit/test_poly.py`:

- `handle_poly_method('concave', ratio=0.3).ratio == 0.3` — regression for the bug
- `handle_poly_method('concave', alpha=0.4).ratio == 0.4` — existing behaviour, now explicit

https://claude.ai/code/session_01QcRmaxVuhvrMsbnxSXdya1

---
_Generated by [Claude Code](https://claude.ai/code/session_01QcRmaxVuhvrMsbnxSXdya1)_

### DIFF
--- a/landau/poly.py
+++ b/landau/poly.py
@@ -518,7 +518,7 @@ with ImportAlarm("'fast-tsp' package required for FastTsp.  Install from pip.") 
 def handle_poly_method(poly_method, **kwargs):
     '''Uniform handling of poly_method between plot_phase_diagram and plot_mu_phase_diagram.
     Some **kwargs trickery required to handle now deprecated min_c_width and alpha arguments.'''
-    ratio = kwargs.pop('alpha', Concave.ratio)
+    ratio = kwargs.pop('alpha', kwargs.pop('ratio', Concave.ratio))
     allowed = {
                 'concave': Concave(**kwargs, ratio=ratio),
                 'segments': Segments(**kwargs),
--- a/landau/poly.py
+++ b/landau/poly.py
@@ -518,7 +518,7 @@ with ImportAlarm("'fast-tsp' package required for FastTsp.  Install from pip.") 
 def handle_poly_method(poly_method, **kwargs):
     '''Uniform handling of poly_method between plot_phase_diagram and plot_mu_phase_diagram.
     Some **kwargs trickery required to handle now deprecated min_c_width and alpha arguments.'''
-    ratio = kwargs.pop('alpha', kwargs.pop('ratio', Concave.ratio))
+    ratio = kwargs.pop('ratio', kwargs.pop('alpha', Concave.ratio))
     allowed = {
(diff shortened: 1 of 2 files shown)


pmrv (commented):
If both ratio and alpha are passed, the deprecated keyword should have *lowest* precedence.

Issue #177, opened 2026-06-02
Plotting Artifacts for unstable phases in 1d_T plots: Return of #9 & #11


pmrv:
<img width="1323" height="756" alt="Image" src="https://github.com/user-attachments/assets/17b31e4f-8ce6-4677-ac7c-b6edda2e2ff1" />

pmrv:
@claude Investigate.  Add this kind of plot with more than 2 stable phases to the testplot script.  For the fix work analogously to #145 i.e. use the cluster_* functions to assign unique labels then pass these as `units=` to seaborn.

pmrv:
Found a related issue in excess f plot (see rectangle highlight in screenshot below)

<img width="883" height="846" alt="Image" src="https://github.com/user-attachments/assets/4e442297-2036-47ad-97ba-3f9075ca2062" />

@claude probably a good time to pull out this kind of stable/unstable labeling from all 1d plots, write hypothesis unit tests and fix this for good.

PR #178, opened 2026-06-02
fix: split disconnected unstable segments in plot_1d_T_phase_diagram

pmrv:
Fix plotting artifact for unstable phases in 1D T plots (#177).

In a three-phase sequence (e.g. bcc/fcc/liquid), the middle phase is unstable in two disjoint temperature ranges. `sns.lineplot` was connecting both into one continuous dashed line, crossing the stable region.

Cluster each (phase, stable) group by T with `cluster_T_c(distance_threshold=0.1)` and pass the result as `units=` to seaborn, following the same approach used in `plot_excess_free_energy` (#145).

Also adds a `plot_1d_T_three_stable` testplot that exposes this scenario.

### DIFF
--- a/landau/plot.py
+++ b/landau/plot.py
@@ -398,11 +398,30 @@ def plot_1d_T_phase_diagram(
 
     if ax is None:
         fig, ax = plt.subplots()
+
+    df = df.copy()
+    # Assign per-segment IDs so that seaborn draws disconnected (phase, stable)
+    # blocks separately.  Without this a phase that is unstable in two disjoint
+    # T ranges (e.g. the middle phase in a three-phase sequence) gets connected
+    # into one continuous dashed line crossing the stable region.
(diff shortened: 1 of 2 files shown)


pmrv:
@claude regression in 1d_T_phase_diagram you lost the stable branches at high and low temperature.  Also the two 1d_T_phase_diagram examples turn out to be equivalent.  Replace them with new example that has some curvature in the free energies so that the bug we're looking at here is better visible.

pmrv:
@claude free energies cannot be concave up, so keep the lines all monotonously decreasing.  Do not come up with new ad hoc clustering methods. try to pick a informed distance threshold for the existing methods based on the data coming into the plot.

pmrv:
plot_1d_T can be discarded then.

pmrv:
> plot_1d_T can be discarded then.

@claude

Issue #34, opened 2025-10-21
Consider generalizing plot_{mu,}_phase_diagram functions


pmrv:
Axes should given as arguments, rather than duplicating almost identical methods.

PR #179, opened 2026-06-02
fix(plot): segment 1d phase diagrams by stability flips, not distance clustering

pmrv:
Fix segmentation artifacts for unstable phases in the 1d phase diagrams (#177).

`plot_1d_mu_phase_diagram` and `plot_1d_T_phase_diagram` split each phase into separate seaborn line segments (via `units=`) so a phase that is unstable in two disjoint ranges draws as separate dashed branches rather than one line crossing its stable window.

### Why not distance clustering

Earlier fixes segmented by a `cluster_T_c` distance threshold; neither a constant nor a data-informed threshold works robustly:

- A constant threshold over-splits a contiguous line on coarse grids once the normalised point spacing exceeds it, shattering the line into one segment per point.
- A data-informed `1.5x`-spacing threshold under-splits genuinely disjoint branches on coarse grids, where the inter-branch gap and the intra-run spacing are comparable, and on the `mu` axis, where concentration is non-linear and a narrow stable c-window (e.g. a liquid saturating near c=1) inflates the threshold above the real branch gap.

Telling a contiguous run from disjoint runs along a 1d scan is an ordering property, not a metric one.

### Approach

Both call sites use `_assign_segment_ids(df, scan_col)`: ordering by the cut axis, a phase begins a new segment at each `stable` flip. This is threshold-free, independent of grid density and of how derived quantities vary along the cut, and parameterised by the cut column (`scan_col`) so it extends unchanged to generalised 1d cuts along arbitrary T/mu/... variables (#34, #60). The now-unused `_auto_threshold` / `_max_normed_gap` helpers are dropped.

`plot_excess_free_energy` keeps `cluster_T_c`: within a constant-T facet it splits disjoint c-ranges of equal-stability rows, where there is no scan order to key off.

### Tests

`tests/unit/plot/test_1d_plots.py` is rewritten around the `scan_col` signature with coarse-grid, density-independence, constant-concentration, and run-count-invariant cases that the threshold approaches failed, plus a coarse 8-point T fixture alongside the fine one.

Closes the 1d segmentation artifacts tracked in #177.


pmrv (commented):
@claude for the 1d mu plot (new in testplot) also used curved semigrand potentials, you can start with the ones from the existing mu plot with 2 phases (replace that test plot actually).

pmrv:
@claude rebase on main

pmrv:
@claude in the mu 1d example fcc phase needs to be shifted slightly lower to ensure three stable phases

pmrv:
@claude the bug (of connecting the two unstable branches) is still present in the mu 1d plot.    first write a test that extracts lines plotted to the axis and ensures there are three distinct line for the fcc phase, both here and in the 1d T case as well.  The same assertion must be present on the segment ids in the existing test cases as well.

pmrv:
@claude do *NOT* roll ad-hoc new clustering code

pmrv:
Conclusion on the 1d-plot segmentation: settled on scan-order ("line scan") rather than distance clustering. Ordering points along the cut axis, a phase starts a new line segment at each `stable` flip — `_assign_segment_ids(df, scan_col)`.

Pragmatic reason: no distance threshold is robust here. A constant `cluster_T_c` threshold over-splits a contiguous line on coarse grids; a data-informed `1.5×`-spacing threshold under-splits genuinely disjoint branches on coarse grids and on the non-linear µ axis (a narrow stable c-window inflates the threshold). Telling a contiguous run from disjoint runs along a 1d scan is an ordering property, not a metric one — scan-order encodes that directly, so it is correct at any grid density and needs no threshold.

Future extension: `scan_col` is a parameter, so the same helper covers generalized 1d diagrams along arbitrary T/µ/… cuts — pass whichever column orders the points. Clustering is kept where it remains the right tool (2d diagrams, excess-free-energy facets).

PR #191, opened 2026-06-03
fix(plot): show legend when only line phases are stable

### LINKED ISSUE #182: plot_excess missing legend if only line phases are stable
Image

pmrv:
When `g._legend` is `None` (no solution-phase hue legend from seaborn), the line-phase handles in `plot_excess_free_energy` were silently dropped. Fall through to empty `existing_handles`/`existing_labels` so the figure legend is always created when line phases are present.

Closes #182.


pmrv:
@claude add such a graph also to testpolot

pmrv:
@testplot

PR #197, opened 2026-06-07
Tune 1d phase-diagram side-label stack: ylim, adaptive width, overlap avoidance

pmrv:
Tunes the 1d phase-diagram side-label legend (`plot_1d_mu_phase_diagram` / `plot_1d_T_phase_diagram`).

## Changes

1. **`ylim` keyword** on both functions. Applied via `ax.set_ylim` like the matplotlib builtin, and used to bound the side-label stack (label `y` is clamped to the window).

2. **Adaptive width reservation.** The room reserved for the right-hand label stack is now derived from the widest *rendered* label, measured in pixels via the figure renderer (`Text.get_window_extent`), not a string-length heuristic. The right x-limit is widened by exactly the fraction of the axes width the widest label occupies.

3. **Vertical overlap avoidance.** Labels in a stack are spread apart so their rendered boxes never overlap (`_spread_labels`, a pure two-pass placer resolved in display space and clamped to `[lo, hi]`).

4. **Left-hand stack for clipped labels.** When `ylim` would clip a label off the top — its line end sits above the window — that phase is labelled at its left line end on a mirrored left-hand stack, reserved and laid out by the same rules. With the default (autoscaled) `ylim` nothing is clipped, so all labels stay on the right and behaviour is unchanged.

## Tests

`tests/unit/plot/test_1d_plots.py`:
- `_spread_labels` unit tests: separation of coincident targets, untouched separated targets, input-order preservation, top-bound respect.
- `ylim` sets axes limits and clamps side labels within the window.
- side labels' rendered boxes fit inside the axes; longer phase names reserve strictly more space than short ones.
- side labels in a stack do not overlap vertically (checked from rendered boxes).
- a tight `ylim` moves clipped phases to the left (`ha='right'`) stack and widens the left x-limit to fit them.

Full file: 65 tests pass; `tests/unit/plot/` 112 pass.

---
_Generated by [Claude Code](https://claude.ai/code/session_01PUP94FpY3n1dCisbGfjKHK)_

### DIFF
--- a/landau/plot.py
+++ b/landau/plot.py
@@ -826,17 +826,20 @@ def plot_1d_mu_phase_diagram(
     if 'border' not in df.columns:
         return ax
 
-    dfa = np.ptp(df['phi'].dropna())
     dfm = np.ptp(df['mu'].dropna())
 
     if mark_transitions:
+        # Label y in axes fraction (blended transform) so a zoomed ylim can't fling the
+        # text out of view; the marker dot stays in data coords and is simply clipped if
+        # the crossing lies outside the window.
         for mt, dd in df.query("mu.min()<mu<mu.max() and border").groupby("mu"):
             ft = dd['phi'].iloc[0]
             ax.axvline(mt, color='k', linestyle='dotted', alpha=.5)
             ax.scatter(mt, ft, marker='o', c='k', zorder=10)
 
-            ax.text(mt - .05 * dfm, ft - dfa * .1, rf"$\Delta\mu = {mt:.03f}\,\mathrm{{eV}}$",
-                    rotation='vertical', ha='center', va='top')
+            ax.text(mt - .05 * dfm, 0.02, rf"$\Delta\mu = {mt:.03f}\,\mathrm{{eV}}$",
+                    transform=ax.get_xaxis_transform(),
+                    rotation='vertical', ha='center', va='bottom')
     ax.set_xlabel("Chemical Potential Difference [eV]")
     ylabel = "Semi-grandcanonical Potential [eV/atom]"
     if reference_phase is not None:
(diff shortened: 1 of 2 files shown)


landau/plot.py
```diff
@@ -534,6 +668,7 @@ def plot_1d_T_phase_diagram(
         reference_phase=None,
         top_labels=True,
         side_labels=True,
+        ylim=None,
```
pmrv, line 869: `ylim` should also be taken as a single float, in which case it is equivalent to (None, ylim).

pmrv: Done in de04ad9 — a scalar `ylim` is treated as `(None, ylim)` (top only, bottom autoscaled), with a test pinning that the bottom stays at its autoscaled value.

---

pmrv (commented):
check whether an applied ylim would drop a phase line entirely from the visible figure, and if so drop the label.

pmrv:
Done in b2867db. When an applied `ylim` pushes a phase's whole line out of view it now gets no label. A phase still visible at its left end but clipped off the top continues to move to the left stack. Visibility is `_phase_visible_in_band` (any sampled point inside the window, or a segment straddling it), with direct unit tests plus a `plot_1d_T` case asserting the dropped phases get no label.

---

pmrv:
experimentally the overlapped adjusted side labels do not seem to respect their original ordering (same as line terminals), fix that.

pmrv:
Fixed in a4a5ad1. The spread now uses the true line-terminal pixels instead of values pre-clamped to the window. Pre-clamping tied any terminals lying outside the window, so their order fell back to the groupby (alphabetical) order; `_spread_labels` already bounds the result to the axes box, so dropping the pre-clamp keeps the labels in view while preserving the terminal order. Added a regression test (`test_side_labels_keep_terminal_order_when_clamped`) that fails on the old path and passes now.

---

pmrv:
ylim has some unfortunate interactions with the transition marker labels. when zoom with ylim, those texts can end up far outside the axis.  Easiest fix might be to plot these after applying ylim in axis coordinates.

pmrv:
Fixed in e42b630. The transition-marker text is now placed in axes-fraction y (blended `get_xaxis_transform`) near the bottom of the axis, instead of at the crossing phi offset by the full data range, so a zoomed `ylim` keeps it in view. The marker dot stays in data coords and is simply clipped when the crossing falls outside the window. Added `test_transition_labels_stay_inside_axis_under_ylim` (mu + T) asserting the label boxes stay within the axes; verified it fails on the old data-coordinate placement.

---

pmrv:
both label side stacks should have a maximum y position so as to not intrude into the space taken for the bold stable phase labels.

pmrv:
Done in 36b8560. Both side stacks now spread only up to just below the rendered bottom of the bold top stable-phase labels (the ceiling is measured from those label boxes and passed into the spread), so neither stack intrudes into that band. Added `test_side_labels_stay_below_top_labels`, verified it fails when the ceiling is the axes top instead of the top-label band.

---

PR #206, opened 2026-06-08
feat(plot): inline phase labels on excess-free-energy facets

pmrv:
Carries the inline-label approach from the recent 2d (`_add_inline_polygon_labels`) and 1d (`_place_side_labels` / `_add_1d_phase_legend`) work over to `plot_excess_free_energy`, which was the last plot still using a side figure-legend box.

## Changes
- New `_add_inline_curve_labels(ax, entries)` helper: draws one label per phase just off its line, white-outlined (`_text_with_outline`) and colour-matched, with mathtext subscripts bolded via `_bold_math` and a high zorder so labels sit above scatter dots and hull markers. Overlapping labels are spread apart vertically via the existing `_spread_labels`.
- Placement keeps labels off the data line: free-energy curves are convex, so solution-phase labels go above the centre of the phase's largest continuous c-region, and line-phase labels go below their dot (line phases sit on the lower hull). The anchor uses the centre of the region's c-span, not the density-weighted point mean (mu-uniform sampling pulls that into the dilute corners). Anchoring on the largest continuous region keeps the label on the prominent stretch rather than on a dilute-corner sliver where such a phase may be the only stable region.
- New `inline_legend=True` parameter matching `plot_phase_diagram`'s API. Default-on drops the figure legend in favour of inline labels; `inline_legend=False` keeps the previous legend box unchanged.

Reused existing helpers (`_text_with_outline`, `_spread_labels`, `_bold_math`, `_get_renderer`) rather than adding new placement code.

## Tests
- Updated `test_all_line_phases_legend_present` to opt into the legend with `inline_legend=False`.
- Added: inline labels on by default drop the figure legend; every facet carries one label per phase (solution + line); label colours match the palette/override; `inline_legend=False` keeps the figure legend and adds no inline text.
- `pytest tests/unit/plot/` — 162 passed.

https://claude.ai/code/session_0112AqtraRTyiacwVFzXwcdy

### DIFF
--- a/landau/plot.py
+++ b/landau/plot.py
@@ -1026,6 +1026,40 @@ def plot_1d_T_phase_diagram(
 # Excess free energy plot
 # ---------------------------------------------------------------------------
 
+def _curve_obstacles(ax):
+    """Drawn curves and scatter markers of *ax* as one shapely geometry in pixels.
+
+    Phase curves become :class:`shapely.LineString`\\ s and scatter dots (line-phase
+    markers, hull vertices) become :class:`shapely.Point`\\ s buffered by their
+    marker radius, all in display coordinates so a label's pixel bounding box can be
+    tested against them directly.  The horizontal reference line (a blended-transform
+    ``axhline``) is skipped so labels are not pushed off the zero line.  Returns the
+    unioned geometry, or ``None`` when nothing is drawn yet.
+    """
+    geoms = []
+    for line in ax.lines:
+        if line.get_transform() is not ax.transData:  # skip refline / blended artists
+            continue
+        disp = line.get_transform().transform(line.get_xydata())
+        disp = disp[np.isfinite(disp).all(axis=1)]
+        if len(disp) >= 2:
+            geoms.append(shapely.LineString(disp))
+    dpi = ax.figure.dpi
+    for coll in ax.collections:
+        offsets = np.asarray(coll.get_offsets(), dtype=float)
+        if offsets.size == 0:
+            continue
+        disp = coll.get_offset_transform().transform(offsets)
+        sizes = coll.get_sizes()
+        for i, (px, py) in enumerate(disp):
+            if not (np.isfinite(px) and np.isfinite(py)):
+                continue
+            s = sizes[i % len(sizes)] if len(sizes) else 36.0
+            radius = max((np.sqrt(s) / 2.0) * dpi / 72.0, 1.0)  # points^2 -> pixel radius
+            geoms.append(shapely.Point(px, py).buffer(radius))
+    return shapely.union_all(geoms) if geoms else None
+
+
 def _add_inline_curve_labels(ax, entries):
     """Label curves/points just off the data line instead of via a legend box.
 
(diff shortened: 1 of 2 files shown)


pmrv:
not happy with the horizontal placement of the labels yet.  If multiple curves span the whole range (ie. almost always) then your current approach will bunch a lot of labels in the center of the plot.  I still think that the initial position of the label should at least be within the extend of the longest continuous arc.

pmrv:
`b394fea`: the x-anchor is no longer the arc centre. The arcs are ordered by their centre and each phase's label is placed at its rank fraction `(i+0.5)/n` of *its own* longest continuous arc, so the anchors fan out left-to-right while each stays within that arc's extent. Curves all spanning [0,1] now get labels at ~0.17/0.5/0.83 instead of all at 0.5. Render in the next testplot comment.

---

landau/plot.py
```diff
@@ -1180,8 +1306,51 @@ def plot_excess_free_energy(
...
+            for i, (pname, cs, fs, color) in enumerate(arcs):
+                frac = (i + 0.5) / n
+                x = cs[0] + frac * (cs[-1] - cs[0])
+                y = np.interp(x, cs, fs)
+                entries.append((pname, x, y, color, "above"))
+            for _, row in sub_lp.drop_duplicates("phase").iterrows():
+                entries.append((row["phase"], row["c"], row["f_excess"], palette.get(row["phase"], "k"), "below"))
+            facet_entries.append((ax, entries))
+
+    if inline_legend:
```
pmrv, line 1356: avoid branching on inline_legend twice, rather try to turn off seaborn from the get go.

pmrv:
`419e33d`: the rank fraction is now mapped into each arc's inner 60% (20% margin per side), so the fanned-out anchor lands on the arc body rather than the turning-up edge — `sigma` now sits over its bump. Render in the next testplot comment.

---

pmrv:
@claude in a new PR critically evaluate code complexity of this PR and try to compress and clean things up a bit.

pmrv (commented):
@claude comment on feasibility

pmrv:
> **Claude finished @pmrv's task in 1m 11s** —— [View job](https://github.com/eisenforschung/landau/actions/runs/27177587905)
> ### Feasibility: turn off seaborn's legend up front
> 
> Feasible and a net simplification. The second `if inline_legend` block (lines 1356–1359) only exists to _undo_ a legend seaborn already built:
> 
> ```python
> if inline_legend:
>     if g._legend is not None:
>         g._legend.remove()
> elif not df_lp.empty:
>     ...augment legend with line-phase handles...
> ```
> 
> `sns.relplot` accepts a `legend` kwarg, so the legend never needs to be created in the inline case. Pass it through at the `relplot` call (line 1247):
> 
> ```python
> g = sns.relplot(
>     ...,
>     legend=False if inline_legend else "auto",
>     ...
> )
> ```
> 
> Then the build-then-remove block collapses to just the augment arm:
> 
> ```python
> if not inline_legend and not df_lp.empty:
>     ...augment legend with line-phase handles...
> ```
> 
> That removes one of the two `inline_legend` branches and the wasted legend construction.
> 
> **Caveats, both benign:**
> 
>     * The empty-`base_data` path (line 1238) builds a bare `FacetGrid` that never makes a legend, so `legend=False` doesn't reach it — but it doesn't need to (`g._legend` is `None` there today, and the removed code was already guarded by `is not None`).
> 
>     * The remaining `inline_legend` branch is the per-facet entry computation at line 1314. That one is irreducible — label anchors must be computed inside the facet loop after the curves are drawn — so it stays. The maintainer's "twice" reduces to "once", which is the unavoidable one.
> 
> 
> Behaviour is identical: in the inline case nothing visible changes (the legend was discarded anyway); in the `inline_legend=False` case the seaborn legend is still created (`legend="auto"`) and augmented exactly as before.
> 
> One thing to verify when implementing: that `legend="auto"` (vs the implicit default) reproduces the current non-inline legend exactly — `relplot`'s default _is_ `"auto"`, so passing it explicitly is a no-op, but worth a glance at the rendered `inline_legend=False` plot to be sure.
> 
> No code changed — this is a feasibility note only, as asked. • branch [`claude/pr-206-20260609-0117`](https://github.com/eisenforschung/landau/tree/claude/pr-206-20260609-0117)

@claude PR away.

PR #214, opened 2026-06-09
Remove dead arctan2 sort in Segments._sort_segments

### LINKED ISSUE #209: Remove dead arctan2 sort in Segments._sort_segments
Refs #116.

## Context

`Segments._sort_segments` in `landau/poly.py:202-258` does an initial sort of its PCA-projected segments by angle around their joint center of mass:

```python
# initial sorting by center of mass angle
segments = sorted(
        segments,
        key=lambda s: np.arctan2( (s[y_col].mean() - com[y_col]) / norm[1],
                                  (s[x_col].mean() - com[x_col]) / norm[0])
)

ordered = _greedy_stitch(segments, norm, x_col, y_col)
```

The first thing `_greedy_stitch` (`landau/poly.py:333`) does is:

```python
head, *remaining = sorted(segments, key=lambda s: s[x_col].min())
```

Python's `sorted` is stable, so the COM-angle order only survives for segments that tie on `min(x_col)`. In refined phase-diagram output those are exact-equal floating-point mins, which essentially never happens — the COM-angle sort is dead code in practice.

## Proposal

Two options, both equally fine:

1. **Delete** the COM-angle sort (4 lines + comment). The `_greedy_stitch` head-picking heuristic is fully responsible for the visible order.
2. **Document** the intent — promote the COM-angle key into a one-line tie-break comment on the `_greedy_stitch` `sorted(... key=s[x_col].min())` call so a future maintainer doesn't reinvent it.

`#162` (the `_sort_segments` rewire) was closed deferred because the next major release is expected to consolidate poly-methods, so option 1 is the lower-friction choice.

## Why

Misleading code: a reader who skims `_sort_segments` will assume the angle order influences output and may spend time understanding the heuristic.

## Acceptance

- The dead block removed (or annotated, per choice above).
- One added test in `tests/unit/test_poly.py` that calls `Segments._sort_segments` directly with three segments arranged so their COM-angle order and their `min(x_col)` order disagree, and asserts the output is ordered by `min(x_col)`. Pins the live heuristic so the choice survives the upcoming poly-method consolidation.

pmrv:
Closes #209.

Removes the dead COM-angle pre-sort in `Segments._sort_segments`. Experiment on the toy and basics refined diagrams confirms removing it gives byte-identical polygons, while the `min(x_col)` head pick in `_greedy_stitch` is live (head choice changes non-convex polygons) and stays. Adds a test pinning the live `min(x_col)` heuristic.

### DIFF
--- a/landau/poly.py
+++ b/landau/poly.py
@@ -226,7 +226,6 @@ class Segments(AbstractPolyMethod):
         if df.empty:
             return pd.DataFrame(columns=df.columns)
 
-        com = df[[x_col, y_col]].mean()
         norm = np.ptp(df[[x_col, y_col]], axis=0).values
         norm = np.where(norm == 0, 1, norm)
 
@@ -247,13 +246,9 @@ class Segments(AbstractPolyMethod):
         if not segments:
             return pd.DataFrame(columns=df.columns)
 
-        # initial sorting by center of mass angle
-        segments = sorted(
-                segments,
-                key=lambda s: np.arctan2( (s[y_col].mean() - com[y_col]) / norm[1],
-                                          (s[x_col].mean() - com[x_col]) / norm[0])
-        )
-
+        # _greedy_stitch fixes the order: it picks the min-x_col segment as head and
+        # nearest-neighbour stitches the rest, so the segment order on entry here does
+        # not matter.
         ordered = _greedy_stitch(segments, norm, x_col, y_col)
         return pd.concat(ordered, ignore_index=True)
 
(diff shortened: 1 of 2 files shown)


landau/poly.py
```diff
@@ -247,13 +246,9 @@ def pca_projection(group):
...
-        # initial sorting by center of mass angle
-        segments = sorted(
-                segments,
-                key=lambda s: np.arctan2( (s[y_col].mean() - com[y_col]) / norm[1],
-                                          (s[x_col].mean() - com[x_col]) / norm[0])
-        )
-
+        # _greedy_stitch fixes the order: it picks the min-x_col segment as head and
+        # nearest-neighbour stitches the rest, so the segment order on entry here does
+        # not matter.
```
pmrv:
```suggestion
```

PR #215, opened 2026-06-09
Draw single-temperature excess free energy as a plain lineplot

pmrv:
`plot_excess_free_energy` always built a seaborn `FacetGrid` (via `relplot`), even for a single temperature, producing a one-column grid. For one temperature it now draws onto a single axes with `sns.lineplot` and returns that `Axes`. Multiple temperatures still fan out into a `FacetGrid`.

The per-axis drawing (metastable curves, line-phase dots, common tangents, hull vertices, inline labels) is unchanged. Legend handling and finalization (`refline`/`axhline`, title, axis labels) are resolved against either the FacetGrid's `_legend`/`figure` or the single axes.

Return type is now `Axes` for one temperature, `FacetGrid` for several; the docstring's Returns section is updated to match. Tests in `tests/unit/plot/test_excess_free_energy.py` are updated: single-T cases assert an `Axes` return and access `ax`/`ax.figure` directly; multi-T return-type and overlap tests keep the `FacetGrid` assertions.

`pytest tests/unit/plot tests/regression` passes (179 tests).

---
_Generated by [Claude Code](https://claude.ai/code/session_01GDrDWGr1L6GBZ3SmZDnVVg)_

### DIFF
--- a/landau/plot.py
+++ b/landau/plot.py
@@ -1183,8 +1183,9 @@ def plot_excess_free_energy(
             False, keep the figure legend box on the right.
 
     Returns:
-        seaborn.FacetGrid: FacetGrid with one column per temperature.
-        Access the figure via ``.fig`` and individual axes via ``.axes``.
+        For a single temperature, the :class:`matplotlib.axes.Axes` holding the
+        plain lineplot.  For multiple temperatures, a :class:`seaborn.FacetGrid`
+        with one column per temperature (figure via ``.fig``, axes via ``.axes``).
     """
     import matplotlib.lines as mlines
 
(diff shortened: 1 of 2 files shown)


landau/plot.py
```diff
@@ -1377,12 +1406,18 @@ def plot_excess_free_energy(
...
-    g.set_titles("T = {col_name:.0f} K")
-    g.set(xlabel="Concentration", ylabel="Free Energy of Formation")
+    if g is not None:
+        g.refline(y=0)
+        g.set_titles("T = {col_name:.0f} K")
+        g.set(xlabel="Concentration", ylabel="Free Energy of Formation")
+    else:
+        ax = axes[0]
+        ax.axhline(0, color=".5", linestyle="--")
+        ax.set_title(f"T = {temperatures[0]:.0f} K")
```
pmrv, line 1416: @claude remove in another PR.

PR #217, opened 2026-06-09
fix(calculate): replace single-T f_excess branch with isinstance check

### LINKED ISSUE #160: Pin pandas-3 single-group `groupby.apply` shape with regression tests for `get_transitions` and `AbstractPolyMethod.apply`
## Background

`cluster_phase` (`landau/plot.py`) was bitten by a pandas 3 quirk where `groupby(...).apply(callable_returning_Series)` returns a transposed DataFrame for *single-group* input instead of the row-aligned Series produced for multi-group input. That site is now pinned in `tests/regression/test_single_group_apply.py` and `test_cluster_phase_pandas3.py`.

Two other groupby-apply sites in the codebase use the same pattern and have no equivalent regression test:

1. `landau/calculate.py::get_transitions` — `bdf.groupby(["mu", "T"])[["c", "phase"]].apply(reduce, include_groups=False)` and then `tdf.groupby("transition", group_keys=False).apply(lambda g: cluster_T_c_mu(g, ...), include_groups=False)`. The second one already has a defensive `if isinstance(res, pd.DataFrame): res = res.stack().reset_index(level=0, drop=True)` shim, suggesting the bug has been hit there. A test that exercises the single-group branch would lock it in.
2. `landau/poly.py::AbstractPolyMethod.apply` — `prepare(df).groupby(['phase', 'phase_unit']).apply(self.make, variables=variables, include_groups=False)` returning a Series of shapely geometries. Single-group behaviour should be checked end-to-end.

There is also a third less-obvious site in `landau/calculate.py::calc_phase_diagram`:

```python
fex = pdf.groupby("T", group_keys=False).apply(sub, include_groups=False)
if len(Ts) > 1:
    pdf["f_excess"] = fex
else:
    # thank you pandas, this saved me -10min of my life.
    pdf["f_excess"] = fex.T
```

This is the same bug, worked around by hand. It deserves either a regression test that pins the `len(Ts) == 1` path, or a cleaner fix that drops the special case (e.g. wrap with `pd.Series(..., index=pdf.index)`). Beware that #155 is also targeting `sub` in the same function — keep the changes orthogonal.

## Acceptance

- A regression test under `tests/regression/` (or extending the existing pandas-3 files) that exercises each of the three sites with single-group input and asserts the resulting DataFrame has the expected shape and dtypes.
- No production-code change required for `get_transitions` or `AbstractPolyMethod.apply` unless the test surfaces a real bug.
- For the `f_excess` site, optionally tighten the assignment to `pdf["f_excess"] = pd.Series(...)` so the `.T` branch is no longer load-bearing.

Sub-issue of #116.

pmrv:
Closes #160.

## Changes

`calc_phase_diagram` used a `len(Ts) > 1` guard and an `fex.T` transpose to
work around pandas 3 returning a `DataFrame` for single-group
`groupby("T").apply(sub)` instead of a `Series`. This is the same pandas 3
quirk already fixed in `get_transitions` (`calculate.py:340`) with an
`isinstance(fex, pd.DataFrame)` check.

Replace the `len(Ts)` branch with the same pattern:

```python
fex = pdf.groupby("T", group_keys=False).apply(sub, include_groups=False)
if isinstance(fex, pd.DataFrame):
    fex = fex.stack().reset_index(level=0, drop=True)
pdf["f_excess"] = fex
```

The new `test_calc_phase_diagram_single_T_f_excess` regression test in
`tests/regression/test_single_group_apply.py` pins the single-T path:
it calls `calc_phase_diagram([...], Ts=[700.0], ..., keep_unstable=True)`
and asserts `f_excess` is present, float, and fully non-null.

Full suite: 376 passed, 5 skipped (`pytest`).

https://claude.ai/code/session_014qYKR3a3asKud3MmxuqAkg

---
_Generated by [Claude Code](https://claude.ai/code/session_014qYKR3a3asKud3MmxuqAkg)_

### DIFF
--- a/landau/calculate.py
+++ b/landau/calculate.py
@@ -229,11 +229,10 @@ def calc_phase_diagram(
         return dd.f - (f0 * (1 - dd.c) + f1 * dd.c)
 
     fex = pdf.groupby("T", group_keys=False).apply(sub, include_groups=False)
-    if len(Ts) > 1:
-        pdf["f_excess"] = fex
-    else:
-        # thank you pandas, this saved me -10min of my life.
-        pdf["f_excess"] = fex.T
+    if isinstance(fex, pd.DataFrame):
+        # pandas 3 returns a DataFrame for single-group groupby.apply; undo that.
+        fex = fex.stack().reset_index(level=0, drop=True)
+    pdf["f_excess"] = fex
     if not keep_unstable:
         pdf = pdf.query("stable")
     return pdf
--- a/tests/regression/test_single_group_apply.py
+++ b/tests/regression/test_single_group_apply.py
@@ -1,14 +1,16 @@
-"""Regression tests for pandas 2/3 single-group groupby.apply (issue #134).
+"""Regression tests for pandas 2/3 single-group groupby.apply (issue #134, #160).
 
 Sites tested:
   - get_transitions: calculate.py:309 ``isinstance(res, pd.DataFrame)`` branch
   - AbstractPolyMethod.apply: poly.py:99 single-group dropna semantics
+  - calc_phase_diagram f_excess: calculate.py:231 single-T groupby.apply shape
 """
 import numpy as np
 import pandas as pd
 from matplotlib.patches import Polygon as MplPolygon
 
-from landau.calculate import get_transitions
+from landau.calculate import calc_phase_diagram, get_transitions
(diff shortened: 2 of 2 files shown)


pmrv:
@claude check whether this also runs on pandas 2, manually installing that and then running tests.

PR #221, opened 2026-06-10
Recover polygons when make_valid returns a GeometryCollection

pmrv:
Fixes the missing solid phase in the second figure (c-T phase diagram) of the Intermetallics notebook, and the polygon amputation visible in the tsp/segment-tsp testplot renders.

Two layers to the bug:

1. `shapely.make_valid` (default linework algorithm) on a self-intersecting ring can return a `GeometryCollection` of the repaired polygon plus zero-area line artifacts. The repair branch in `AbstractPolyMethod.make` only handled `Polygon`/`MultiPolygon`, so such regions were dropped outright — the solid phase vanished from the Intermetallics c-T diagram. Repair now uses `make_valid(method="structure", keep_collapsed=False)`, which returns the enclosed polygon directly; minimum shapely bumped to 2.1 for the `method` argument.
2. The self-intersections come from `solve_tsp_record_to_record`, which keeps invalid tours often enough to matter (6–12 of 20 runs per phase on the toy mu diagram). Even when repair succeeds, taking the largest part amputates the pinched-off rest of the region. `PythonTsp` now re-solves with a 10x larger iteration budget when the tour polygon is invalid (`retries=2`); `SegmentPythonTsp` falls back to the exact DP solver for endpoint TSPs with at most `exact_node_limit=16` nodes, since iteration growth does not help its random restarts (8 of 12 retry ladders still invalid at 1000 iterations) while the exact tour was valid in every observed run.

Tests:
- `tests/unit/test_poly.py::test_make_repairs_ring_self_intersection` pins the repair: a ring retracing a zero-width spike must come back as a valid polygon with the enclosed area, not `None`.
- `tests/integration/test_border_coverage.py` checks the full `calc_phase_diagram` → `get_polygons` pipeline: every finite stable border point must lie within 2% of the axis ranges of its phase's polygon outline, so amputated regions fail immediately. Parametrized over segments/tsp/segment-tsp/fasttsp/segment-fasttsp × c-T/mu-T; segments × mu-T is xfailed (min-x head-heuristic limitation documented in `Segments._sort_segments`, fails identically on main).

Re-executed Intermetallics notebook committed. Full suite: 392 passed, 1 xfailed.

https://claude.ai/code/session_01PQDckXjinohTr6LiByat3g


pmrv:
bump minimum shapely and use the nicer algo

2d_basics_mu_phase_diagram_segment-tsp and 2d_toy_mu_phase_diagram_tsp and 2d_toy_mu_phase_diagram_segment-tsp are all broken (missing parts of the polygon, so I doubt you actually fixed the issue.  

Include the rerendered intermetallics notebook in commits to ensure you are actually doing things correctly here.

pmrv:
Root cause of the tsp-only breakage: `solve_tsp_record_to_record` keeps self-intersecting tours often enough to matter (6–12 of 20 runs per phase on the toy mu diagram for segment-tsp; reproducible for tsp on basics-mu). The invalid tour then went to repair, where taking the largest part of the `make_valid` output amputated the pinched-off rest of the region. The first commit only stopped the phase from being dropped entirely; it did nothing against amputation.

Changes in 8587fc2:
- `PythonTsp`: re-solve with a 10x larger iteration budget when the tour polygon is invalid (`retries=2`). 100 iterations cured the reproducible basics-mu case.
- `SegmentPythonTsp`: iteration growth does not help here — its restarts are random and 8 of 12 retry ladders were still invalid at 1000 iterations. Instead it falls back to the exact DP solver when the endpoint TSP has ≤16 nodes (the toy/basics instances have 12–14; exact tour was valid in every observed run, <1 s).
- Repair switched to `make_valid(method="structure", keep_collapsed=False)`, shapely minimum bumped to 2.1.
- `tests/integration/test_border_coverage.py` implements the vice-versa check: every finite stable border point must lie within 2% of the axis ranges of its phase's polygon outline, parametrized over segments/tsp/segment-tsp/fasttsp/segment-fasttsp × c-T/mu-T. The segments × mu-T combination fails on main as well (0.495 normalized distance, the documented min-x head-heuristic limitation in `Segments._sort_segments`) and is xfailed.

The three flagged plots render correctly across repeated local runs; the gallery re-render on this push should confirm. Re-executed Intermetallics notebook committed in 2e15834.

---

PR #233, opened 2026-06-11
Move DelaunayTripleRefiner dedup state from solve() into run()

### LINKED ISSUE #210: Move DelaunayTripleRefiner dedup state out of solve() into run()
Refs #116.

## Context

```python
def __init__(self):

    ...
           for fT, fmu in self._found):
        return []
    self._found.append((T, mu))
```

## Proposal

## Acceptance

Sub-issue of #116.

pmrv:
Closes #210. Sub-issue of #116.

## Changes

`landau/refine.py` — `DelaunayTripleRefiner`:

- Remove `__init__` and `self._found`. `solve()` is now a pure function of `(cand, phases)`: it minimizes the triple potential and returns the `RefinedPoint` unconditionally.
- Add a `run()` override. It maintains a local `found: list[TMuPoint]` across candidates and drops any result whose `(T, mu)` falls within the enclosing simplex's extent of an already-emitted point — the same tolerance logic that was previously inside `solve()`. The docstring documents that `run()` owns dedup so future subclassers do not reintroduce it in `solve()`.

`tests/unit/test_refine.py`:

- Add `test_delaunay_triple_solve_is_pure`: calls `solve()` twice on the same candidate and asserts both calls return a non-empty result at the same location.
- `test_delaunay_triple_refiner_deduplicates` and `test_boundary_id_delaunay_triple_rows_share_id` pass unchanged.

Full suite: 397 passed, 0 failed.

https://claude.ai/code/session_01D4u5mwhm3nkZru3DB6KYAs

---
_Generated by [Claude Code](https://claude.ai/code/session_01D4u5mwhm3nkZru3DB6KYAs)_

### DIFF
--- a/landau/refine.py
+++ b/landau/refine.py
@@ -410,12 +410,15 @@ class DelaunayTripleRefiner(Refiner):
     For every Delaunay simplex containing three distinct phases, locate the
     triple point by minimizing the sum of pairwise potential differences,
     starting from the simplex centroid.
+
+    Deduplication is handled by :meth:`run`, not :meth:`solve`. A point
+    returned by ``solve`` is dropped by ``run`` when it falls within the
+    enclosing simplex's ``(T, mu)`` extent of an already-emitted triple
+    point. Subclasses that override ``solve`` do not need to reimplement
+    this logic.
     """
 
     label = "delaunay-triple"
-
-    def __init__(self):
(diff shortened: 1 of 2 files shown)


pmrv:
This makes the refiner the only one that overrides run.

PR #235, opened 2026-06-11
Add locus column classifying phase diagram points

pmrv:
Tag every row of the `calc_phase_diagram` dataframe with a `locus` column: `interior` for plain grid samples, `boundary` for refined two-phase coexistence points and miscibility-gap branches, `triple` for refined three-phase points. Until now triple points were only recognizable via `refined == "delaunay-triple"` or by counting phases per `(T, mu)` group.

Values are backed by the new `Locus` StrEnum in `landau/features.py` (import via `from landau.features import Locus`; not re-exported at the package level yet), so rows select both by plain string and by member:

```python
df.query("locus == 'triple'")
df[df["locus"] == Locus.TRIPLE]
```

Implementation:
- `RefinedPoint.to_rows` tags by the number of coexisting phases (3 → `triple`, else `boundary`), keeping the tagging refiner-agnostic; `RefinedMiscibilityGap.to_rows` tags both branch rows `boundary`.
- Grid samples are tagged `interior` in `calc_phase_diagram` (covers `refine=False`) and in `_split_stable` (covers direct `refine_phase_diagram` calls).
- The synthetic frame rows from `_border_edges` (`mu=±inf`, T extremes) keep `locus = interior` despite `border = True` — they only close the sampling window and are no thermodynamic transitions.

Tests: locus assertions added to the existing ClausiusClapeyron, DelaunayTriple, and miscibility-gap refiner tests plus direct `to_rows` unit tests in `tests/unit/test_refine.py`; `_split_stable`/`_border_edges` and end-to-end `calc_phase_diagram` coverage (unrefined all-interior, refined ⇔ non-interior, frame edges, triple point location) in `tests/unit/test_calculate.py`. Full suite: 410 passed, 1 xfailed on pandas 3.0.3.

https://claude.ai/code/session_01Q38JJsxAV8sErJiYoCFY6p

### DIFF
--- a/landau/__init__.py
+++ b/landau/__init__.py
@@ -1,5 +1,3 @@
-from .features import Locus
-
 from .phases import (
     LinePhase,
     TemperatureDepandantLinePhase,
--- a/landau/__init__.py
+++ b/landau/__init__.py
@@ -1,3 +1,5 @@
+from .features import Locus
+
 from .phases import (
     LinePhase,
     TemperatureDepandantLinePhase,
--- a/tests/unit/test_calculate.py
+++ b/tests/unit/test_calculate.py
@@ -380,7 +380,8 @@ def test_cluster_use_mu_false_dispatches_to_cluster_T_c():
 # --- locus column tests ---
 
 
-def _triple_point_phases():
+@pytest.fixture
+def triple_point_phases():
     """Three LinePhases with a single triple point at (T=300 K, mu=0.2 eV)."""
     return [
         LinePhase("A", 0.0, -1.0, 0.004),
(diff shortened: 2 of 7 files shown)


tests/unit/test_calculate.py
```diff
@@ -367,3 +375,49 @@ def test_cluster_use_mu_false_dispatches_to_cluster_T_c():
     labels = cluster(dd, use_mu=False, distance_threshold=0.5)
     assert labels.nunique() == 1
     assert len(labels) == len(dd)
+
+
+# --- locus column tests ---
+
+
+def _triple_point_phases():
```
pmrv: fixture

landau/__init__.py
pmrv, line 1: don't export globally yet

PR #246, opened 2026-06-15
Add low-temperature expansion point defect sublattice

pmrv:
Adds a low-temperature (dilute) expansion of the point defect model and tidies the surrounding code.

**Model.** `PointDefectSublattice` evaluates the exact per-site semi-grand potential

    omega_l = -kB*T * ln(1 + sum_i z_i),   z_i = exp(-beta [phi_i])

whose denominator encodes site competition (`sum_i x_i <= 1`). The leading term of its
low-temperature expansion replaces `ln(1 + sum_i z_i)` with `sum_i z_i`, dropping site competition
so the defect site fractions `x_i = z_i` are independent and unnormalised. Accurate while defects
are dilute (`z_i << 1`); converges to the exact model there.

**Shared ABC.** Both models derive from `AbstractPointDefectSublattice` (frozen-dataclass ABC)
holding the fields, the shared `_get_zes`, and the `semigrand_potential`/`concentration` combiners.
Each concrete model supplies two hooks — `_site_log_partition` (per-site potential in units of
`-kB*T`) and `_site_normalization` (site-fraction denominator), coupled by `c = -∂φ/∂Δμ ⇒ D = 1/L'`.
Exact: `L = ln(1+s)`, `D = 1+s`. LTE: `L = s`, `D = 1`. The two are **siblings**, not a subclass chain.

**Clamp + tangent extension.** The unnormalised LTE site fractions would drive the total `c` out of
`[0,1]` and `φ → -∞` outside the dilute regime, making an LTE phase spuriously win `argmin φ`. So
`PointDefectedPhase` clamps `c` to `[0,1]` and continues `φ` as the **tangent line** past each
saturation point (a line phase at `c=1`, slope `-1`, or `c=0`, slope `0`). Since `c = -∂φ/∂Δμ`, the
clamped `c` and linear `φ` stay thermodynamically consistent, and the bounded `φ` no longer
dominates the diagram. The saturation chemical potentials depend only on `T`, so they're precomputed
per temperature (`_saturation_window`) inside a non-overflowing Δμ bound; the raw expression is only
evaluated within the window. A no-op for the bounded exact model.

**Module split.** The point defect classes move out of `phases/__init__.py` into a dedicated
`landau/phases/pointdefects.py` (following `asewrapper.py`). The pre-split public names stay
importable from `landau.phases` for back-compat until 2.0 — `AbstractPointDefect` as a plain alias,
the concrete `ConstantPointDefect`/`PointDefectSublattice`/`PointDefectedPhase` via `@deprecate`
shims; the new `AbstractPointDefectSublattice`/`LowTemperatureExpansionSublattice` are imported from
`landau.phases.pointdefects`. Unused diagnostic-only `ConstantPointDefect` methods dropped.

**Docs / diagnostics.** `notebooks/PointDefects.ipynb` gains a *Low-Temperature Expansion* section,
the exact and LTE T–c phase diagrams, a 1D μ diagram via `plot_1d_mu_phase_diagram`, and a scatter
overplot comparing the exact vs LTE phase borders; the B2 example uses slightly asymmetric defect
energies. `benchmarks/point_defect_expansions.py` compares host vs exact vs LTE (raw vs stitched)
across temperatures.

**Tests** (`tests/unit/test_phases.py`): closed-form leading terms (`rtol=1e-12`); full-range
`c = -∂φ/∂Δμ` consistency at sublattice and phase level *including the clamped region*; `c ∈ [0,1]`;
`φ` bounded by the tangent extension; dilute-limit convergence to / divergence from the exact model;
a Hypothesis property test; the sibling-under-ABC structure; and back-compat re-export tests
(annotated for removal at 2.0).

https://claude.ai/code/session_01HvxTR7HQNBqCr1cYq4Eghz


pmrv:
Host vs exact (`PointDefectSublattice`) vs low-temperature expansion (`LowTemperatureExpansionSublattice`) for the B2 model from `notebooks/PointDefects.ipynb`, across temperatures. Top row: concentration `c(Δμ)`; bottom row: defect contribution `φ − φ_host`. Generated by `benchmarks/point_defect_expansions.py`.

point_defect_expansions

The LTE tracks the exact model while defects are dilute and departs once site competition matters: its unnormalised site fractions push `c` past 1 and `φ` plunges, while the exact model stays bounded (`c ∈ [0,1]`, shallow `φ` well). The Δμ half-window over which the two concentrations agree to 0.01 shrinks monotonically with T:

| T [K] | agree \|Δμ\| [eV] |
|------:|------------------:|
|   300 |             0.249 |
|   800 |             0.169 |
|  1500 |             0.063 |
|  3000 |             0.017 |

At 3000 K the defects are no longer dilute even at stoichiometry, so the models diverge across the whole window.

---

pmrv:
Added a clamp: `PointDefectedPhase.concentration` now clips the total to `[0, 1]` via `np.clip`. The LTE drops the site-competition denominator, so its unnormalised site fractions can push `c` past the physical bounds outside the dilute regime; the clamp keeps the reported concentration physical and is a no-op for the bounded exact model. `φ` is left untouched.

Regenerated diagnostic — the dotted line is the raw (pre-clamp) LTE total, the dashed line is the clamped result held at `c = 1`:

point_defect_expansions_clamped

Tests (`tests/unit/test_phases.py`): `test_lte_phase_concentration_clamped_to_physical_range` asserts the clamped phase stays in `[0, 1]` at T=2000 K over `|Δμ| ≤ 1` while the raw sum exceeds 1 and drops below 0; `test_lte_sublattice_in_phase_gibbs_duhem` is narrowed to `|Δμ| ≤ 0.2` where the clamp is inactive so the `c = -∂φ/∂Δμ` identity still holds.

---

pmrv:
Implemented your suggested fix: keep the `c` clamp to `[0,1]` **and** continue `φ` as the tangent line past each saturation point. Since `c = -∂φ/∂Δμ`, a constant clamped `c` (0 or 1) requires φ to be linear there with slope `−c` — i.e. the phase becomes a line phase at the saturated concentration. This restores `c = -∂φ/∂Δμ` consistency *and* bounds φ, so the LTE phase no longer spuriously wins `argmin φ`.

The saturation chemical potentials depend only on `T` (not on the query Δμ), so they're precomputed a priori per temperature (`_saturation_window`, bracketed inside a non-overflowing Δμ bound); the raw, overflow-prone expression is then only evaluated for Δμ *inside* the window.

Result — `AB_LTE` is now a proper compound region (`c ∈ [0.43, 0.65]`) instead of smearing across the whole axis, and there are no overflow warnings:

point_defect_expansions

Dotted = raw unstitched LTE (overshoots `c=1`, φ → −∞); dashed = clamped `c` + tangent φ.

Tests (`tests/unit/test_phases.py`): `c = -∂φ/∂Δμ` now asserted over the **full** Δμ range including the clamped region (`test_lte_phase_phi_c_consistency_full_range`, exercising both saturated branches); `c ∈ [0,1]`; φ bounded by the tangent extension vs the raw `< −100` plunge; plus the Hypothesis property test on the sublattice contributions. The notebook's LTE section and full T–c diagram were re-executed and now render cleanly.

---

landau/phases/__init__.py
pmrv, line 1: Add tests to ensure that the moved classes stay importable, annotate the tests that these are to be slashed at 2.0 release and decorate the rexport classes with the deprecator.

pmrv:
Done in 6774c21. Re-exports from `landau.phases`, removed at 2.0:

- `ConstantPointDefect`, `PointDefectSublattice`, `PointDefectedPhase` → `@deprecate(..., version="2.0")` factory shims pointing at `landau.phases.pointdefects` (factory functions, like `TemperatureDepandantLinePhase`, so the warning names `landau.phases.X` rather than the canonical module).
- `AbstractPointDefect` → plain alias (it predates the split, kept subclassable; `@deprecate` would turn it into a function and break that).
- `AbstractPointDefectSublattice` and `LowTemperatureExpansionSublattice` are new here, so per your note they have no back-compat obligation — not re-exported, dropped from `phases.__all__`; import from `landau.phases.pointdefects`.

Back-compat tests in `tests/unit/test_phases.py` (block annotated for removal at 2.0) cover importability, the ABC-is-canonical alias, the concrete shims warning + constructing the canonical class, and the new classes staying off `landau.phases`. Notebook/test imports moved to the canonical module.

Also added a notebook cell overplotting the stable phase-border points of the exact vs LTE diagrams (frame edges dropped via `locus`) for a direct comparison.

---

pmrv:
increase mu grid size in notebook

pmrv:
rebase on main for merge conflicts, tweak AB defect phase energies such that they are slightly assymetric.

pmrv:
the plot ' K: semi-grand potentials and stable phases' can be replaced by the newer 1D plots

lte phase should be named AB (LTE), reads nicer

pmrv:
Addressed the recent comments:

- Rebased onto `origin/main` (linear history, force-pushed); the only conflict was the `RegularSolution` imports/tests in `test_phases.py`, resolved to keep both sets.
- Defect energies: lowered the α-sublattice antisite (`B_alpha` 0.30→0.25) and vacancy (`V_alpha` 0.50→0.45) below their β counterparts. My earlier 0.35/0.55 actually *cancelled* the pre-existing solid/liquid asymmetry — this goes the other way, so the compound is clearly lopsided toward the B-rich side. The exact-vs-LTE border overlay now shows the LTE over-extending the compound there (where defects stop being dilute).
- `mu` grid increased to 200 (explicit Δμ ranges, which also avoids the exact model's `exp`-overflow region that `guess_mu_range` would otherwise probe).
- Replaced the hand-rolled "semi-grand potentials and stable phases" figure with `plot_1d_mu_phase_diagram`.
- Renamed the phase to `AB (LTE)`.

Notebook re-executes cleanly (no errors/warnings); full unit suite 422 passed, 4 skipped.

---

pmrv:
You're right — reverted to `segment-fasttsp` (the default). Nothing is actually wrong with it; `concave` is the broken one here.

The LTE compound over-extends all the way to **c = 1** (vs the exact model's ~0.73), so the region touches the right corner. Diagnostic, polygon outline over the `AB (LTE)` border points:

poly method diagnostic

- `segment-fasttsp` (left): hugs the border points exactly. The raw segment tour self-intersects at the corner, but `make_valid(method="structure")` (PR #221) recovers the correct ring — that's all the `…produced an invalid polygon; repairing it` warning is reporting.
- `concave` (right): leaves a spurious notch/gap near c=1 and detaches the corner strip — genuinely wrong.

So the diagrams are back on `segment-fasttsp`. The only side effect is that benign repair `UserWarning` in the LTE cell's output (cell 18/exact stays clean since it doesn't reach the corner). Want me to leave it as-is (honest, it's the documented repair path), or wrap that one cell in `warnings.catch_warnings` to keep the output clean?

---

pmrv:
@claude squash and update OP

pmrv:
Heads-up: the automated squash (`cd75f74`) had two regressions, so I redid it as `e8cf8bb`:

- it reverted `origin/main`'s recent commits — `landau/calculate.py` and `tests/unit/test_calculate.py` matched the pre-`b5252b4` state (the branch was squashed without rebasing onto the advanced `main`);
- it dropped `benchmarks/point_defect_expansions.py` (that file is part of this PR, not a net-deleted file).

`e8cf8bb` is the 14 commits squashed into one, rebased onto the current `origin/main` (`459151c`): `calculate.py`/`test_calculate.py` now match `main`, the benchmark is present, and the only `CLAUDE.md` delta vs `main` is the split/LTE/deprecation text. Full suite: 448 passed, 4 skipped. OP updated to match (the benchmark line is kept, since the file is in the diff).

---

PR #247, opened 2026-06-15
Rewrite triple-point lines on the locus column; rename tielines arg

pmrv:
Rewrites the triple-point line drawing in `landau/plot.py` on the `locus` column added in PR #235, renames the plotting argument, and extends the mark to μ-T diagrams.

## Plotting

`_plot_tielines` re-derived "where is a triple point" by counting phases per temperature, in a `refined`/unrefined split that carried `# TODO`/`# FIXME` and a variable-reuse bug. The `locus` column already classifies that: a three-phase invariant emits rows tagged `Locus.TRIPLE` sharing one `(mu, T)`. `_plot_triplepoints` groups those rows by `(mu, T)` and marks each invariant:

- c-T: one isothermal `hlines` (`lw=2`) spanning the coexisting concentrations.
- μ-T: a black point marker at the `(mu, T)` where the three phases collapse to a single point.

Unrefined frames carry no `TRIPLE` rows and draw nothing; a frame without a `locus` column is a no-op. `get_transitions` is no longer used in `plot.py` (import removed).

## Argument

`plot_phase_diagram` and `plot_mu_phase_diagram` gain `triplepoints` (the latter so the μ-T markers are reachable). The old `plot_phase_diagram(tielines=...)` keyword stays as a deprecated argument (`@deprecate(tielines=...)`) that warns and maps onto `triplepoints`.

## Tests

`tests/unit/plot/test_triplepoint.py` is the single home for triple-point tests (replaces `test_tielines.py`, which pinned the deleted heuristic; the per-axis cases previously in `test_polygons.py` are consolidated here). Shared `ax` and `triple_df` fixtures; cases parametrized over the axis pair:

- synthetic geometry — c-T draws one isothermal line per invariant spanning the coexisting concentrations; μ-T draws one black marker per invariant `(mu, T)`; non-`TRIPLE` rows ignored.
- no-op when the `locus` column is absent or no `TRIPLE` rows exist.
- `plot_phase_diagram(tielines=...)` warns and routes to `triplepoints`.
- end-to-end: a real refined eutectic diagram (`eutectic_diagram` fixture) marks the invariant in both c-T and μ-T.

`pytest tests/unit/plot` passes.

https://claude.ai/code/session_01L8wQPeuGiR356hrruzxCn7


landau/plot.py
```diff
@@ -410,27 +386,31 @@ def _plot_phase_diagram(
...
+    tielines="Use plot_triplepoint instead",
 )
 def plot_phase_diagram(
     df,
     alpha=0.1,
     element=None,
     min_c_width=1e-2,
     color_override: dict[str, str] = {},
-    tielines=False,
+    plot_triplepoint=False,
```
pmrv: call this `triplepoints`

pmrv: in mu/T plots this should then put black markers on triple points

tests/unit/plot/test_polygons.py
pmrv, line 1: triple point tests niw spread over mutilple points. Unify and drop duplicate test cases

tests/unit/plot/test_triplepoint.py
```diff
@@ -1,32 +1,137 @@
...
+    assert sorted(_hlines(ax)) == pytest.approx([(300.0, 0.1, 0.9), (450.0, 0.2, 0.7)])
+    assert _markers(ax) == []
+    plt.close(fig)
+
+
+def test_muT_draws_black_marker_per_invariant():
+    """mu-T: one black marker at each invariant (mu, T), no isothermal lines."""
+    fig, ax = plt.subplots()
+    _plot_triplepoints(_triple_df(), ax=ax, variables=["mu", "T"])
+    assert _hlines(ax) == []
```
pmrv: pull into fixture

pmrv (commented):
revise

pmrv (commented):
update OP and try and compress tests without losing coverage.

PR #259, opened 2026-06-20
refactor(refine): attribute triple points to owning simplex; remove _found state

### LINKED ISSUE #210: Move DelaunayTripleRefiner dedup state out of solve() into run()

pmrv:
Closes #210.

`DelaunayTripleRefiner` held a mutable `self._found` list that `solve()` wrote to and read across candidates to dedup triple points found by adjacent three-phase simplices. The list made `solve()` stateful and caused a silent no-op on a second call with the same candidate, contrary to every other refiner.

A Delaunay tessellation partitions space and the triple point is a single point, so the located minimum lies inside exactly one three-phase simplex. The fix attributes the point to its owning simplex via a barycentric containment check (`_point_in_simplex`): `solve()` returns `[]` when the minimum falls outside the candidate simplex and emits the point when it falls inside. No cross-call state is needed.

Changes:
- `_point_in_simplex(T, mu, simplex)` helper added (barycentric, affine-invariant; shared edges counted as inside so a point on a Delaunay edge is claimed by both neighbours rather than neither).
- `DelaunayTripleRefiner.__init__` removed; `self._found` gone.
- `solve()` is now a pure function of `(cand, phases)`.
- Class docstring documents attribution-by-containment so subclassers do not reintroduce state in `solve()`.
- New test `test_delaunay_triple_solve_is_pure_and_simplex_owned`: asserts exactly one candidate owns the triple point and that re-calling `solve()` on the owner returns the same result.
- Existing `test_delaunay_triple_refiner_deduplicates` unchanged and green.

`pytest tests/unit/test_refine.py tests/regression tests/integration/test_border_coverage.py` → 47 passed, 8 skipped, 1 xfailed.

---
_Generated by [Claude Code](https://claude.ai/code/session_017TAxvkNdZM4LZPfZrdjKxZ)_

### DIFF
--- a/landau/refine.py
+++ b/landau/refine.py
@@ -434,17 +434,36 @@ class DelaunayLineRefiner(Refiner):
         return [RefinedPoint(T=T, mu=mu, phases=(name1, name2))]
 
 
+def _point_in_simplex(T: float, mu: float, simplex: pd.DataFrame) -> bool:
+    """Return True if ``(T, mu)`` lies inside or on the boundary of the triangle.
+
+    Uses barycentric coordinates: solve ``A @ lam = p - v0`` for ``lam``, then
+    check ``lam >= 0`` and ``sum(lam) <= 1``.  Shared edges are counted as
+    inside (``>=`` / ``<=`` rather than strict), so a triple point that lands
+    exactly on a Delaunay edge is claimed by both neighbours rather than
+    neither — a duplicate is safer than a miss.
+    """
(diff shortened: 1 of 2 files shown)


pmrv:
@claude diagnose CI failure here, then fix

PR #274, opened 2026-06-22
feat(interpolate): add SplineFit concentration interpolator

pmrv:
Adds `SplineFit`, a `ConcentrationInterpolator` wrapping `scipy.interpolate.UnivariateSpline`.

## Flags

- `smoothing` (scipy's `s`): the strict-interpolation / smoothing control.
  - `0.0` (default): strict interpolation — the spline passes through every sample.
  - `> 0`: least-squares smoothing — the spline may miss the samples by a total squared residual of `smoothing`.
  - `None`: defers to scipy's default (`s` equal to the number of samples).
- `degree` (scipy's `k`, default 3 = cubic): clamped to `len(x) - 1` so a fit with fewer samples than `degree + 1` drops to the highest degree the data supports (e.g. two points → linear).

`fit` sorts the input by `x` and returns the usual scalar-in/scalar-out callable (0-d result collapsed to a Python float; array input returns an array). Extrapolation extends the edge segment (scipy's default `ext=0`).

Exported from `landau.interpolate` and re-exported from the package root, matching `PolyFit` / `SoftplusFit`.

## Tests

`tests/unit/interpolate/test_spline_fit.py` (9 tests), sharing a single `ATOL = 1e-9`:

- `smoothing=0` reproduces every sample to `ATOL`.
- Hypothesis round-trip: a cubic interpolating spline recovers any degree-≤3 polynomial at dense interior points, not just at the nodes (a degree-`k` spline reproduces degree-`k` polynomials to roundoff; worst error over 2000 random ±10 cubics was 1.4e-14).
- Shuffled input interpolates identically (internal sort).
- Scalar in → `float`; array in → array of matching shape.
- `degree=1` matches `np.interp` — plain piecewise-linear interpolation.
- Degree clamp: two points fit a line through them.
- `smoothing > 0` departs from the noisy samples up to the requested residual budget and lands closer to the underlying clean signal than the strict fit.
- `smoothing=None` smooths (does not reproduce the nodes).
- Frozen-dataclass hashability/immutability (usable as a cache key).

Full suite: 511 passed, 1 xfailed locally (502 existing + 9 new).

https://claude.ai/code/session_01AUXQZRmrsSZCRkK4uy2mPF

---
_Generated by [Claude Code](https://claude.ai/code/session_01AUXQZRmrsSZCRkK4uy2mPF)_


landau/interpolate/basic.py
```diff
@@ -100,6 +101,46 @@ def fit(self, x, y):
...
+        order = np.argsort(x)
+        x = x[order]
+        y = y[order]
+        k = min(self.degree, len(x) - 1)
+        spline = UnivariateSpline(x, y, k=k, s=self.smoothing)
+
+        def interpolation(c):
+            f = spline(c)
+            if np.ndim(c) == 0:
+                return f.item()
```
pmrv: we have a helper for this, iirc.

pmrv (commented):
revise

landau/interpolate/basic.py
pmrv: Done in aac1206. `_scalarize` lived in `landau.phases`, which imports `interpolate`, so pulling it the other way would cycle — moved the definition down into `landau/interpolate/basic.py` (phases re-imports it; `pointdefects` import unchanged) and `SplineFit` now uses it. Left the older inline copies in `G_calphad`/`StitchedFit` as-is to keep this PR focused.

---

pmrv (commented):
@claude rebase to fix conflicts

pmrv:
Rebased onto `439ef9a`. The only content conflict was in `landau/phases/__init__.py`: #279 removed `Sprime`/`c_from_dmu` while my refactor removed the local `_scalarize` def in favour of the `from ..interpolate.basic import _scalarize` import — both deletions combined, so the whole block between `S(c)` and `class Phase` is gone. Full suite green locally (511 passed, 1 xfailed). Updated the PR body for the post-rebase counts (9 new tests).

---

PR #277, opened 2026-06-22
refine: attribute triple points to owning simplex instead of dedup state

### LINKED ISSUE #210: Move DelaunayTripleRefiner dedup state out of solve() into run()

pmrv:
Closes #210.

`DelaunayTripleRefiner` held a mutable `self._found` list that `solve()` wrote to and read across candidates to dedup triple points found by adjacent three-phase simplices. A Delaunay triangulation partitions space and the triple point is a single point, so the located minimum lies inside exactly one three-phase simplex. `solve()` now emits only when the minimum falls within its own candidate simplex, making it a pure function of `(cand, phases)`: no `__init__`, no `_found`, no `run` override.

### DIFF
--- a/landau/refine.py
+++ b/landau/refine.py
@@ -474,9 +474,7 @@ class DelaunayTripleRefiner(Refiner):
     triple point by minimizing the sum of pairwise potential differences,
     starting from the simplex centroid.
 
-    Near a physical triple point the tessellation may produce several adjacent
-    three-phase simplices, each seeding the same minimisation. The located
-    minimum may fall outside every three-phase simplex — inside a
+    The located minimum may fall outside every three-phase simplex---inside a
     neighbouring two-phase triangle (the Delaunay partition has at most one
     triangle that strictly contains any point). Ownership is therefore
     determined by :func:`_simplex_containment`: the sibling with the largest
--- a/landau/refine.py
+++ b/landau/refine.py
@@ -474,15 +474,17 @@ class DelaunayTripleRefiner(Refiner):
     triple point by minimizing the sum of pairwise potential differences,
     starting from the simplex centroid.
 
-    Near a physical triple point the tessellation produces several adjacent
-    three-phase simplices, each of which seeds the same minimum. The point is
-    attributed to the single simplex that owns it — the one whose triangle
-    contains it, or, when the coarse grid leaves the minimum just past a shared
-    edge, the one it is least far outside of (largest :func:`_simplex_containment`
-    score over the sibling simplices). :meth:`solve` emits only when its own
-    simplex is that owner, so exactly one candidate fires. The siblings travel
-    in the candidate, keeping :meth:`solve` a pure function of ``(cand, phases)``
-    — no ``__init__``, no dedup state, no ``run`` override.
+    Near a physical triple point the tessellation may produce several adjacent
+    three-phase simplices, each seeding the same minimisation. The located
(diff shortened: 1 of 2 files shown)


pmrv:
@claude fix ci

pmrv:
@claude[sonnet] your prose is still wrong. There can only be one delauney triangle that contains a triple point (at most 2, or 3 if it were to fall exactly on a shared edge or corner, but unlikely). More likely is that the triple point is outside the tri-colored simplex that is used to seed the refinement.

pmrv:
@claude your prose is still wrong. There can only be one delauney triangle that contains a triple point (at most 2, or 3 if it were to fall exactly on a shared edge or corner, but unlikely). More likely is that the triple point is outside the tri-colored simplex that is used to seed the refinement.

pmrv:
@claude use softer language: it *may* fall outside the original triangle

landau/refine.py
```diff
@@ -439,25 +473,32 @@ class DelaunayTripleRefiner(Refiner):
...
+    Near a physical triple point the tessellation may produce several adjacent
+    three-phase simplices, each seeding the same minimisation. The located
+    minimum may fall outside every three-phase simplex — inside a
+    neighbouring two-phase triangle (the Delaunay partition has at most one
+    triangle that strictly contains any point). Ownership is therefore
+    determined by :func:`_simplex_containment`: the sibling with the largest
+    score (least far outside the minimum) is the owner. :meth:`solve` emits
+    only when its own simplex is that owner, so exactly one candidate fires per
+    triple point. The siblings travel in the candidate, keeping :meth:`solve` a
+    pure function of ``(cand, phases)`` — no ``__init__``, no dedup state, no
```
pmrv:
```suggestion
    The located minimum may fall outside every three-phase simplex---inside a
    neighbouring two-phase triangle (the Delaunay partition has at most one
    triangle that strictly contains any point). Ownership is therefore
    determined by :func:`_simplex_containment`: the sibling with the largest
    score (least far outside the minimum) is the owner. :meth:`solve` emits
    only when its own simplex is that owner, so exactly one candidate fires per
    triple point. The siblings travel in the candidate, keeping :meth:`solve` a
    pure function of ``(cand, phases)`` — no ``__init__``, no dedup state, no
```

landau/refine.py
```diff
@@ -439,25 +473,30 @@ class DelaunayTripleRefiner(Refiner):
...
+
+    The located minimum may fall outside every three-phase simplex---inside a
+    neighbouring two-phase triangle (the Delaunay partition has at most one
+    triangle that strictly contains any point). Ownership is therefore
+    determined by :func:`_simplex_containment`: the sibling with the largest
+    score (least far outside the minimum) is the owner. :meth:`solve` emits
+    only when its own simplex is that owner, so exactly one candidate fires per
+    triple point. The siblings travel in the candidate, keeping :meth:`solve` a
+    pure function of ``(cand, phases)`` — no ``__init__``, no dedup state, no
+    ``run`` override.
```
pmrv:
```suggestion
    pure function of ``(cand, phases)`` --- no ``__init__`` or no dedup state.
```

Issue #138, opened 2026-05-19
Thermodynamic database to landau layer


prabhath-c:
Plot .tdb files with landau

prabhath-c:
@claude What are your thoughts on creating a function to read phase diagrams from thermodynamic database files (.tdb) from calphad like methods into landau and plot it using landau? How feasible would this be?

prabhath-c:
@pmrv What are your thoughts on adding pycalphad as an optional dependency? If not, I can always have a local function (outside of landau) to do this for me.

pmrv:
@claude Sketch a CEF phase class.  Restrict to the binary case and the `free_energy` method only. Conversion into semigrand ensemble can be done generically later.

pmrv:
> [@pmrv](https://github.com/pmrv) What are your thoughts on adding pycalphad as an optional dependency? If not, I can always have a local function (outside of landau) to do this for me.

Let's do whatever necessary for a prototype for now.  I'm not as convinced as claude that just parsing tdb files is so much harder than calling pycalphad, but it's definitely something we can tackle in a second step.

pmrv:
@prabhath-c @claude  I found [equilipy](https://github.com/ORNL/Equilipy/) which looks like the tool we need.

pmrv:
AH, it doesn't read tdb though :/

pmrv:
If it's only for line phases it's sunk for our case anyway.

prabhath-c:
> AH, it doesn't read tdb though :/

@pmrv Ah, that's quite unfortunate. But regardless, this would still me we need to add an external package as an optional dependency. Are we okay with doing that?

pmrv:
@prabhath-c  yeah, I think it's fine. optional deps don't hurt and we can later trim if it turns out not to work.

pmrv:
[...]

PR #281, opened 2026-06-23
feat: FastInterpolatingPhase, a vectorized exact replacement for SlowInterpolatingPhase

pmrv:
## Summary

Adds `FastInterpolatingPhase`, a vectorized solver that computes the same
semigrand minimization as `SlowInterpolatingPhase` —
`phi(dmu) = min_c [ fe(c) - T*S(c) - dmu*c ]` — without the per-`dmu`
`scipy.optimize.brute` call. `SlowInterpolatingPhase` is left untouched as the
reference oracle.

As preparation, fitted 1-D interpolations gain a `deriv()` method
(`landau.interpolate`), so the phase solver can take an analytic first
derivative instead of finite differences.

## Motivation

`SlowInterpolatingPhase._find_phi_c_scalar` runs one `brute` (20-point grid +
Nelder-Mead polish) per scalar `dmu`, rebuilding an `np.vectorize` wrapper on
every objective evaluation. On the representative drive (scalar-`T` loop over a
120-point `dmu` array) that is ~1.4 s per phase. Its 20-point grid can also miss
a deep, narrow well and return a non-global minimum.

## Approach

For a fixed `T` the free-energy curve `f(c) = fe(c) - T*S(c)` is built once:

1. one vectorized grid `argmin` over `dmu` locates the global basin (handles
   miscibility gaps);
2. the minimum is polished by a few Newton steps in the logit variable
   `u = log(c/(1-c))`. The ideal-mixing entropy contributes
   `-T*S'(c) = kB*T*u`, which is *linear* in `u`, so the polish is uniformly
   conditioned from the dilute to the saturated limit;
3. the polish is confined to the grid cell around the basin, and the lowest of
   {Newton result, cell edges} is kept, so a boundary minimum is recovered
   exactly and the global basin is never abandoned.

Using `fe.deriv()` (analytic for PolyFit/RedlichKister) makes the located `c`
the exact stationary point, so `c = -d(phi)/d(dmu)` (Gibbs-Duhem) holds to
machine precision. A bytes-keyed result cache lets `semigrand_potential` and
`concentration`, called separately by `calc_phase_diagram` with the same
`(T, dmu)`, solve once.

## `landau.interpolate` change

`Interpolation` becomes an ABC: callable `f(x)` plus `deriv()` returning `f'` as
another `Interpolation`. The default `deriv()` is a scale-aware vectorized
central difference; closed forms override it — `PolyFit` returns a
`PolynomialInterpolation` (poly1d wrapped; coefficients exposed in ascending
power order), `SGTE` returns an `SGTEInterpolation` with the analytic `dG/dT`,
and `RedlichKister` derives from `_eval_mix_derivative`. Stitched/Softplus/
Whitney closures are wrapped so they inherit the numerical default.

## Numbers

From `benchmarks/bench_fast_interpolating_phase.py` (drive: 14 temperatures x
120 `dmu`; error is max |dphi| against a 60001-point dense-grid minimum):

| phase | slow | fast | speedup | slow err | fast err |
|---|---|---|---|---|---|
| PolyFit interior | 1484 ms | 11.7 ms | 127x | 1.5e-8 | 6.8e-11 |
| RedlichKister | 1351 ms | 9.8 ms | 138x | 4.1e-6 | 7.8e-7 |
| double-well | 1706 ms | 9.7 ms | 175x | 4.8e-6 | 8.1e-7 |

Over a 24k-point random-phase sweep the fast solver matched the dense-grid
minimum to max |dphi| = 3.2e-6, while `brute` had 25-33 points where its coarse
grid landed in the wrong basin (max |dphi| = 0.60). Interior stationarity
residual `|f'(c) - dmu|` is 2e-15 (PolyFit) / ~1e-9 (RedlichKister).

## Tests

- `tests/unit/test_phases.py`: shape contracts, true-minimum reproduction across
  PolyFit/RedlichKister and saturation, deep-well global-basin recovery,
  `c = -d(phi)/d(dmu)` consistency (convex phase + exact stationary-point check),
  and agreement with `SlowInterpolatingPhase` where `brute` is reliable.
- `FastInterpolatingPhase` subclasses `SlowInterpolatingPhase`, so it inherits
  the `concentration_range` / `maximum_extrapolation` mutual exclusion from #278;
  a Fast-specific test pins that.
- `tests/unit/interpolate/test_polyfit.py`: coefficient recovery now checked
  through the ascending-order `.coefficients` interface.
- `tests/integration/testplots.py`: the `excess_free_energy` render uses
  `FastInterpolatingPhase`, and the regular-solution liquid in `2d_toy` /
  `2d_toy_mu` is built via `FastInterpolatingPhase` too (same Redlich-Kister
  fit, figures unchanged), so `testplot` exercises the new solver across the
  excess-energy, 2D c-T and 2D T-mu renders.

Full unit + regression suite passes (the 6 pre-existing `test_poly.py` failures
are a local shapely<2.1 environment issue, unrelated to this change).


pmrv:
### How the solver works, in words

Each call has to evaluate, for a fixed temperature `T` and many chemical
potentials `μ`, the semigrand potential

```
φ(μ) = min over c of   f(c) − μ·c ,   with   f(c) = fe(c) − T·S(c)
```

where `fe` is the fitted interpolation and `S(c) = −kB[c ln c + (1−c) ln(1−c)]`
is the ideal mixing entropy. `SlowInterpolatingPhase` solves this with one
`scipy.optimize.brute` per `μ`: a 20-point scan of `c` followed by a
Nelder-Mead polish, and a fresh `np.vectorize` wrapper rebuilt on every function
evaluation. That is robust but slow, and the 20-point scan occasionally settles
in the wrong basin when the landscape has a deep, narrow well.

**The reformulation.** At the optimum the derivative vanishes, `f'(c) = μ`, i.e.

```
fe'(c) − T·S'(c) = μ .
```

The entropy derivative is `S'(c) = −kB·logit(c)`, where `logit(c) = ln(c/(1−c))`.
So if we change variables to `u = logit(c)` (with the inverse `c = sigmoid(u)`),
the entropy contribution to the stationarity condition becomes

```
fe'(c) + kB·T·u = μ ,
```

which is **linear in `u`**. The awkward, logarithmically stiff part of the
problem near the dilute (`c → 0`) and saturated (`c → 1`) limits — exactly where
a Newton step in `c` blows up — turns into a straight line in `u`. Newton's
method in `u` is then uniformly well behaved across the whole composition range.

**The algorithm**, for one temperature:

1. Evaluate `f(c)` once on a grid spanning the concentration range. A single
   vectorized `argmin` over `μ` (broadcast against that one curve) picks, for
   every `μ` at once, the grid cell containing the global minimum. Doing this on
   a curve that is shared across all `μ` is where most of the speed comes from —
   the slow version re-walked the curve for each `μ` separately.
2. Polish with a few Newton iterations in `u`, **confined to the grid cell**
   around the basin the grid picked. Confining it is what stops Newton from
   wandering off to a different stationary point on a non-convex (e.g. PolyFit)
   landscape — the failure mode the plain logit-Newton had before this guard.
3. Take the lowest of {polished point, the two cell edges}. If the true minimum
   sits exactly on a range boundary, that boundary is recovered exactly instead
   of being chased by Newton.

**Why concentration and the gradient of φ stay consistent.** By the envelope
theorem, `∂φ/∂μ = −c` *at the exact stationary point*. Using the analytic
`fe.deriv()` (this PR also gives the interpolators a `deriv()` method) makes the
located `c` the exact stationary point — the residual `|f'(c) − μ|` is ~1e-15 for
polynomials — so the returned `c` equals `−∂φ/∂μ` to machine precision rather
than to a finite-difference tolerance. (Across a genuine miscibility gap `c`
jumps and `φ` has a kink; that is physics, and the relation holds within each
branch.)

**Caching.** `calc_phase_diagram` asks for `semigrand_potential` and
`concentration` separately at the same `(T, μ)`. A small bytes-keyed cache makes
the second call reuse the first solve instead of repeating it.

Net effect on the representative phases: same answer as `brute` where `brute` is
right, the *correct* answer where its coarse grid was wrong, and ~2 orders of
magnitude faster. The numbers and the script that produced them are in the PR
description and `benchmarks/bench_fast_interpolating_phase.py`.

— posted by Claude (Claude Code), on Marvin's account

pmrv:
@prabhath-c — flagging this in case it helps your work. `FastInterpolatingPhase`
is a drop-in for `SlowInterpolatingPhase` (same constructor, same results) that
is ~2 orders of magnitude faster, so if you build interpolating phases from
database/TDB data (cf. #138) or sweep many temperatures and chemical potentials,
swapping the class should speed things up without changing the numbers. It also
finds the global minimum on deep/narrow wells where the old `brute` grid could
land in the wrong basin. No action needed on your side — just thought it might
be useful.

— Claude (Claude Code), posting on Marvin's account

landau/interpolate/basic.py
```diff
@@ -30,8 +30,72 @@ def G_calphad(T, pl, *p):
...
+    """A fitted 1-D interpolation ``f(x)``.
+
+    Subclasses must be callable. :meth:`deriv` returns ``f'`` as another
+    :class:`Interpolation`; the default is a vectorised central difference,
+    and subclasses with a closed form (polynomials, SGTE, Redlich-Kister)
+    override it with the exact derivative.
+    """
+
+    @abstractmethod
+    def __call__(self, x):
```
pmrv: type hints

landau/interpolate/basic.py
```diff
@@ -30,8 +30,72 @@ def G_calphad(T, pl, *p):
...
+
+
+class NumericalDerivative(Interpolation):
+    """``f'`` by a scale-aware central difference; the generic fallback derivative."""
+
+    def __init__(self, func: Callable, step: float = 1e-6):
+        self.func = func
+        self.step = step
+
+    def __call__(self, x):
```
pmrv: try https://docs.scipy.org/doc/scipy/reference/generated/scipy.differentiate.derivative.html#scipy.differentiate.derivative

landau/phases/__init__.py
```diff
@@ -647,6 +648,127 @@ def check_concentration_interpolation(
...
+        cl = conc[np.maximum(idx - 1, 0)]
+        cr = conc[np.minimum(idx + 1, self._n_grid - 1)]
+        with np.errstate(divide="ignore"):
+            # -inf/+inf at the 0/1 ends keep the dilute/saturated tails reachable
+            ua = np.where(cl <= 0.0, -np.inf, se.logit(np.clip(cl, 1e-300, 1.0)))
+            ub = np.where(cr >= 1.0, np.inf, se.logit(np.clip(cr, 0.0, 1.0 - 1e-16)))
+
+        u = np.clip(se.logit(np.clip(conc[idx], 1e-15, 1.0 - 1e-15)), ua, ub)
+        h2 = self._fd2
+        for _ in range(self._n_newton):
```
pmrv, line 720: scipy doesn't have something equally performant?

landau/phases/__init__.py
```diff
@@ -647,6 +648,127 @@ def check_concentration_interpolation(
...
+            c = se.expit(u)
+            fp = fe_prime(c)
+            fpp = (fe(c + h2) - 2 * fe(c) + fe(c - h2)) / (h2 * h2)
+            g = fp + kT * u - flat                  # stationarity residual f'(c) = dmu
+            gp = fpp * (c * (1.0 - c)) + kT          # dg/du
+            u = np.clip(u - g / gp, ua, ub)
+        c_newton = np.clip(se.expit(u), a, b)
+
+        # keep the lowest objective among the polished point and the cell edges
+        # (fe is called on 1-D arrays only: RedlichKister evaluation uses np.vander)
```
pmrv: outdated?

pmrv (commented):
@claude revise

Needs docstrings.

Keep benchmarking as you go, do not allow performance regressions, report performance when you're done.

pmrv:
@claude resolve conflicts by rebase

prabhath-c:
@pmrv Thank you so much! Will definitely use this and let you know.

PR #288, opened 2026-06-26
fix: uncross segment-TSP ring at the c=1 corner

### LINKED ISSUE #253: segment-fasttsp emits a self-intersection repair warning for phase regions touching the c=1 corner
`SegmentFastTsp` builds a self-intersecting ring for a stable region that reaches the `c = 1` corner of the c–T diagram. The ring is recovered correctly by `make_valid(method="structure")` (PR #221), so the resulting polygon is right, but the repair path emits a `UserWarning`:

```
landau/poly.py:69: UserWarning: SegmentFastTsp._make produced an invalid polygon
(Self-intersection[-3.38... 1.15...]); repairing it.
```

The outline is correct after repair (it hugs the border points); `concave` is *not* a workaround here — it leaves a spurious notch/gap at the corner. So the fix should be in the segment-TSP reconstruction (avoid the self-intersection at the corner so no repair/warning is needed), not a poly-method switch. Not urgent.

### Reproducer

```python
import warnings, numpy as np
from scipy.constants import Boltzmann, eV
from landau.phases import LinePhase, IdealSolution
from landau.phases.pointdefects import (
    ConstantPointDefect, LowTemperatureExpansionSublattice, PointDefectedPhase,
)
from landau.calculate import calc_phase_diagram
from landau.plot import get_polygons

kB = Boltzmann / eV

# point-defected compound whose (LTE) stable region extends all the way to c = 1
host = LinePhase("AB", 0.5, -0.40, 1.0 * kB)
lte = PointDefectedPhase("AB (LTE)", host, [
    LowTemperatureExpansionSublattice("alpha", 0, 0.5, [
        ConstantPointDefect("B_alpha", 0.25, 0, +1),
        ConstantPointDefect("V_alpha", 0.45, 0, 0)]),
    LowTemperatureExpansionSublattice("beta", 1, 0.5, [
        ConstantPointDefect("A_beta", 0.30, 0, -1),
        ConstantPointDefect("V_beta", 0.50, 0, 0)]),
])
solid = IdealSolution("solid", LinePhase("A", 0, -0.33, 1.0 * kB), LinePhase("B", 1, -0.30, 1.0 * kB))
liquid = IdealSolution("liquid", LinePhase("A(l)", 0, -0.10, 3.3 * kB), LinePhase("B(l)", 1, -0.13, 3.3 * kB))

df = calc_phase_diagram([solid, liquid, lte], Ts=np.linspace(200, 1400, 25), mu=np.linspace(-0.8, 0.8, 200))

with warnings.catch_warnings(record=True) as w:
    warnings.simplefilter("always")
    get_polygons(df, poly_method="segment-fasttsp")

hits = [str(x.message) for x in w if "invalid polygon" in str(x.message)]
assert hits, "expected the self-intersection repair warning"
print(hits[0])
```

prints

```
SegmentFastTsp._make produced an invalid polygon (Self-intersection[...]); repairing it.
```

### Context

Surfaced in PR #246 (`AB (LTE)` compound over-extends to `c = 1`). Diagnostic — `segment-fasttsp` (correct) vs `concave` (gap at the corner) over the region's border points:

poly method diagnostic

pmrv:
Closes #253.

A phase region touching the c=1 corner splits its boundary into separate segments, forcing a straight closing chord on the opposite side that cuts across a convex nose whose PCA-terminal point is not its geometric extreme — so the ring self-intersects even with an optimal tour. `_segment_tsp_polygon` now uncrosses the assembled ring (2-opt) when invalid, keeping every border point and returning a simple polygon so no make_valid repair/warning is needed. Non-segment poly methods unchanged.

### DIFF
--- a/landau/poly.py
+++ b/landau/poly.py
@@ -411,7 +411,54 @@ def _segment_tsp_polygon(
     coords = np.vstack(chunks)
     if len(coords) < 3:
         return None
-    return shapely.Polygon(coords)
+    polygon = shapely.Polygon(coords)
+    if not polygon.is_valid:
+        polygon = shapely.Polygon(_uncross_ring(coords))
+    return polygon
+
+
+def _segments_cross(p1: np.ndarray, p2: np.ndarray, p3: np.ndarray, p4: np.ndarray) -> bool:
+    """Whether the open segments ``p1 p2`` and ``p3 p4`` properly cross.
+
+    Shared endpoints or collinear touching do not count -- only a genuine
+    interior crossing of the two segments.
+    """
+    def ccw(a, b, c):
+        return (b[0] - a[0]) * (c[1] - a[1]) - (b[1] - a[1]) * (c[0] - a[0])
+
+    d1, d2 = ccw(p3, p4, p1), ccw(p3, p4, p2)
+    d3, d4 = ccw(p1, p2, p3), ccw(p1, p2, p4)
+    return ((d1 > 0) != (d2 > 0)) and ((d3 > 0) != (d4 > 0))
+
+
+def _uncross_ring(coords: np.ndarray) -> np.ndarray:
+    """Remove self-intersections from a closed ring of points by 2-opt reversal.
+
+    The segment-endpoint TSP minimises connecting distance but does not
+    guarantee a simple polygon: where a segment's PCA-ordered terminal point is
+    not its geometric extreme -- a curved nose, e.g. the left side of a phase
+    region whose opposite side touches the ``c=1`` corner -- the closing chord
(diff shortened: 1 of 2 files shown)


pmrv:
truly topologically weird fields should be forbidden by the thermodynamics, so I'm guessing this here is more caused by the clustering artifacts or interplay between different boundaries coming together.

I want to test at least one option where we polish a segmenttsp run with a short pass of the full tsp.

pmrv:
@claude [opus] test that last suggestion of mine

pmrv:
Let's keep this in backup in case it comes up again.

Issue #22, opened 2025-09-19
Add tests


pmrv:
- [ ] Solution phases should be tested against the slow interpolation phase
- [ ] Temperature dependent line phases should be checked in their interpolation accuracy
- [ ] all phase should be tested to accept scalar and numpy vectorized inputs

pmrv:
@claude check current tests and report gaps back.

pmrv:
@claude Build tests for TemperatureDependentLinePhase and IdealSolution.

pmrv:
fastinterpolatingphase is the new standard

PR #292, opened 2026-06-29
test(phases): cross-validate RegularSolution against SlowInterpolatingPhase

pmrv:
Adds two cross-validation tests for issue #22: `test_regular_solution_vs_slow_semigrand_potential` and `test_regular_solution_vs_slow_concentration`.

Both `RegularSolution` and `SlowInterpolatingPhase` fit the same `RedlichKister` polynomial to `(line_fe + T*S)` at the sampled concentrations and expose the same `semigrand_potential`/`concentration` interface. The only implementation difference is the minimisation back-end: `RegularSolution` uses a dense convex-hull loop while `SlowInterpolatingPhase` uses `brute+fmin`. For the same 3-phase fixture (`_make_interp_line_phases()`, 1 RK parameter each), both produce results agreeing to `atol=_INTERP_ATOL = 5e-3` over an 11-point `dmu` sweep at `T = 1000 K`. In practice the difference is ~1e-9 (effectively the same algorithm).

The existing `test_interpolating_vs_slow_*` pair (PR #265) already covers `InterpolatingPhase` against `SlowInterpolatingPhase`. This PR adds the `RegularSolution` side, which was the remaining cross-validation gap flagged in issue #22.

---
_Generated by [Claude Code](https://claude.ai/code/session_01D1dEQvbcx3WR5smxJmPp4J)_

### DIFF
--- a/tests/unit/test_phases.py
+++ b/tests/unit/test_phases.py
@@ -606,6 +606,36 @@ def test_interpolating_vs_slow_concentration():
     assert_allclose(fast.concentration(1000.0, dmu), slow.concentration(1000.0, dmu), atol=_INTERP_ATOL)
 
 
+def test_regular_solution_vs_slow_semigrand_potential():
+    """RegularSolution and SlowInterpolatingPhase agree on semigrand_potential.
+
+    Both fit the same RedlichKister polynomial to (line_fe + T*S) at the sample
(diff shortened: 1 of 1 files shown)


pmrv:
we'll drop both phases in the future and no code uses RegularSolution, so meh.

PR #293, opened 2026-06-29
test(refine): direct unit tests for _dominated

pmrv:
Refs #116.

`_dominated(pt, phases)` is the predicate every refiner's `run()` uses to drop
a refined transition whose phases are not globally stable at `(pt.T, pt.mu)`:
True iff some phase outside `pt.phase_names()` has a strictly lower phi there.
The integration path is exercised by `test_scan_refiner_splits_dominated_crossing`
(narrow B window in mu / T), but the helper itself has no direct test, so a
regression in the filter would only surface as a downstream miscount.

Seven new cases in `tests/unit/test_refine.py`, each pinning one orthogonal
branch of the contract:

- no rivals: empty generator → False
- outside rival strictly below: True
- outside rival strictly above: False
- outside rival exactly equal: False (strict `<` lets a true triple point survive)
- size-3 candidate whose only would-be dominator is *in* its own set: False (the
  `p.name not in own` filter holds)
- size-3 candidate with a fourth phase outside that dominates: True
- two rivals (one higher, one lower): True (lower one alone is enough)

Fixture is a small `LinePhase` mapping built from `(name, c, E)` tuples, so each
test reads as one line of setup plus one assertion.

`pytest tests/unit/test_refine.py` — 34 passed.

---
_Generated by [Claude Code](https://claude.ai/code/session_01QE8BDqF5RYseeVbsBUvfjE)_

### DIFF
--- a/tests/unit/test_refine.py
+++ b/tests/unit/test_refine.py
@@ -692,40 +692,40 @@ def _dom_phases(*specs):
 
 
 def test_dominated_no_rivals_returns_false():
-    # Two phases coexist, the mapping carries only those two: nothing else can
-    # dominate. The generator is empty and any() returns False.
+    """Two phases coexist, the mapping carries only those two: nothing else
+    can dominate. The generator is empty and any() returns False."""
     phases = _dom_phases(("A", 0.0, 0.0), ("B", 1.0, 0.0))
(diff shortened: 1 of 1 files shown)


pmrv (commented):
@claude turn comments into docstrings

PR #298, opened 2026-06-29
Native 2D Phases

pmrv:
Based on FastInterpolatingPhase, but forcing a global fit instead separate for each temperature.  This enforces regularization across temperatures and might improve performance somewhat, since we now fit once and apply many times, rather than the previous delayed fitting per temp at phase diagram construction time.

Ships two adapted interpolators that make use of this:
1. Calphad-style SGTE-terminals + Redlich-Kister
2. Softplus 

1 for solution phases, 2 works nice for intermetallics.


pmrv:
the usual 'getting stuck' issue still happens for the softmax one.  should think about whether we can transform the input variables somehow to make the fit easier.

pmrv:
@prabhath-c The 2D softplus fitter is now much better at finding the correct-ish minimum in a V-shaped free energy landscape.  It should also be faster than separate fits.  I'll have this open for a while for tuning, but let me know if you have feedback.

pmrv:
superseeded for technical reasons in #321

PR #300, opened 2026-06-29
fix(interpolate): robustify SoftplusSurface2D fit (slope in 1/T, convex-well seeding)

pmrv:
Stacked on #298 (`feat/calphad-phase`). The 2-D softplus surface fit got stuck in bad minima on V-shaped wells whose sharpness varies with temperature: it placed the knee and branch slopes badly, worst at the temperature extremes. Two changes.

## Slope basis: `b` is now a polynomial in 1/T

A convex well sharpens as it cools, so the softplus slope `b` scales with inverse temperature. `b_i(T)` is now a polynomial in normalised 1/T; amplitude, knee and offset stay in T. A low-order polynomial in T cannot track the sharpening over a wide T-range, one in 1/T can.

`SoftplusFittedSurface` gains the 1/T normalisation (`wm, ws`) and evaluates `b` from it. `_model`/`_jac` take the 1/T Vandermonde for the `b` columns; the analytic Jacobian is finite-difference-checked (worst |analytic − FD| = 1.2e-9).

## Seeding: convex-well-aware, replacing the central-slice multistart

The coupled fit is multimodal — sharp V-wells sit in a flat-gradient landscape that traps a single optimisation. Each coupled start is now a per-slice seed (`_fit_slice`), with the knee and the branch slopes read from *different* features of the data:

- **knee** = the curvature centre, found as the `argmin` of the data with its best-fit line in c removed (`_knee_position`). A softplus term's knee is where its curvature concentrates; a raw `argmin` instead returns the function *minimum*, which for an asymmetric well sits off the knee and, when the entropy-removed free energy is a small dimple riding a chemical-potential tilt, collapses onto the downhill window edge.
- **branch amplitudes** = one-sided line fits, but the data is split at the function *minimum* (its turning point), not at the knee. The two softplus arms are monotone only on either side of where `H` turns around, so splitting at the minimum keeps each line fit on a monotone arm even when the minimum is off the knee.

The pair is seeded as an opposite-slope softplus (the smooth-V identity `softplus(bu)+softplus(−bu) ~ |u|`) and polished with bounded `least_squares`, `x_scale='jac'` and a 2-stage sharpness ramp. The coldest and central slices seed the coupled fit, so seeding cost is independent of how many temperatures are sampled (hundreds in practice). The coupled solve runs with `x_scale='jac'` plus a tiny ridge toward the seed, which pins flat directions so a redundant term's slope/knee can't run away. `loss` defaults to `linear` — a robust loss would down-weight the well bottom, the signal here, not an outlier.

## Evidence

`benchmarks/bench_softplus_surface_minima.py` fits V-shaped wells (per-branch slope and curvature varying with T, ~120 temperatures) with the slope basis toggled, and reports per-slice normalised RMSE (mean / 95th pct) and worst knee error:

```
surface          Trange    |  b~poly(T)  m/p95/kerr |  b~poly(1/T) m/p95/kerr
sharp_asym       300-900   |  0.009  0.019  0.0042  |  0.008  0.018  0.0040
wide_asym        150-1500  |  0.118  0.261  0.0401  |  0.054  0.122  0.0120
infamily_1overT  150-1500  |  0.033  0.086  0.0180  |  0.004  0.007  0.0172
```

Over a narrow T-range the basis barely matters; over a wide one the poly-in-T slope misfits at the extremes (wide-T asymmetric V: p95 0.26 → 0.12; in-family well that sharpens as 1/T: 0.09 → 0.007). The public class recovers an exactly in-family `b∝1/T` surface to machine precision.

## Tests

`tests/unit/interpolate/test_softplus_surface.py` adds: exact recovery of a `b∝1/T` surface (`atol=1e-7`), accuracy at every temperature on a strong-T-sharpening V (per-slice nRMSE < 1e-3, knee error < 5e-3), and the knee debias under a linear tilt. The existing recovery / convexity / analytic-derivative / Gibbs–Duhem cases are unchanged. Full suite: 545 passed, 15 skipped, 1 xfailed.


landau/interpolate/softplus.py
```diff
@@ -210,27 +211,159 @@ def dfdc(cc):
...
+        self._Tm, self._Ts, self._wm, self._ws = Tm, Ts, wm, ws
+        self._cm, self._cs = cm, cs
 
     def slice_at(self, T: float) -> _SoftplusSlice:
         Tn = (float(T) - self._Tm) / self._Ts
+        Wn = (1.0 / float(T) - self._wm) / self._ws
         pv = np.polynomial.polynomial.polyval
         a = _softplus(np.array([pv(Tn, row) for row in self._A]))  # non-negative amplitude
-        b = np.array([pv(Tn, row) for row in self._B])
+        b = np.array([pv(Wn, row) for row in self._B])             # slope: polynomial in 1/T
```
pmrv, line 353:
```suggestion
        b = np.array([pv(Wn, row) for row in self._B]) 
```

landau/interpolate/softplus.py
```diff
@@ -397,26 +565,84 @@ def fit(self, T, c, f) -> SoftplusFittedSurface:
 
         Tm, Ts = float(T.mean()), float(T.std() or 1.0)
         cm, cs = float(c.mean()), float(c.std() or 1.0)
+        w = 1.0 / T
+        wm, ws = float(w.mean()), float(w.std() or 1.0)
         Tn = (T - Tm) / Ts
+        Wn = (w - wm) / ws
```
pmrv: is normalizing T, then W really required or helpful at all?  We are later fitting polynomials in W, which presumably do the same normalization

pmrv: The polynomial fit doesn't normalize anything — `np.vander(Wn, …)`/`polyval` build raw powers of whatever variable they're handed. Without the centering/scaling the slope basis is powers of raw 1/T ∈ [6e-4, 2.5e-3] over 400–1600 K, and that Vandermonde is badly conditioned: cond ≈ 5e5 / 3e8 / 2e11 at `b_order` 2 / 3 / 4, vs 5 / 25 / 130 normalized. Same reason T is normalized for the a/c/offset polys (raw T⁴ ~ 1e12).

It's mathematically redundant — a poly in raw 1/T spans the same space — but the solve is finite precision. `x_scale='jac'` (and lm's internal scaling) only partly compensate: on a noise-free in-family b∝1/T surface the normalized fit reaches machine precision (nRMSE ~1e-10, cost ~4e-14) while raw 1/T plateaus ~4 orders short (nRMSE ~8e-7, cost ~2e-8) — the high-order slope terms never engage. So I'd keep it.

---

PR #301, opened 2026-06-30
perf(interpolate): speed up SoftplusSurface2D fit

pmrv:
The 2-D softplus surface fit spent its time running two full coupled fits (one per seed, keeping the cheaper) and grinding the per-slice seed fits to `max_nfev` on sharp slices. This cuts the redundant work and picks the coupled solver by scipy version.

Changes (the fitted surface is unchanged):

- **Cap the per-slice seed fits** (`_SEED_MAX_NFEV`). A seed only has to locate the basin; past a few hundred iterations the extra polish does not help, but a hard slice would otherwise grind to `max_nfev`.
- **Race the two seeds** (coldest + central) instead of converging both. Each gets a short polish; the better-converging one is finished with a full solve **restarted from its seed** (not the trial point), so it re-treads the bounded single-fit trajectory and structureless data can't drift to a degenerate basin.
- **Fuse residual + Jacobian** into one shared `softplus`/`sigmoid` pass (`_model_and_jac`) behind a one-slot cache.
- **Pick the coupled solver by scipy version** (`_LM_HAS_INTERNAL_SCALING`). scipy 1.16 ([gh-22790](https://github.com/scipy/scipy/pull/22790), fixing [gh-19459](https://github.com/scipy/scipy/issues/19459)) made `least_squares(method='lm')` use MINPACK's internal variable scaling (`mode=1`), as plain `leastsq` always did — before 1.16 it silently disabled it (`mode=2`, all-ones `diag`) and converged poorly on the badly-scaled sharp-knee Jacobian. So on **scipy ≥ 1.16** the solve uses `lm` (no per-iteration SVD → faster); on **older scipy** it falls back to the trust-region solver with `x_scale='jac'`. The minimum-deps floor (`scipy>=1.11.2`) is unchanged.
- **New `method` field** (`'lm'` / `'trf'`, default `None` = the version-based pick above) to override the solver explicitly. A robust `loss` always uses `trf`; `method='lm'` with a robust loss raises in `__post_init__`.

Fit wall-time at 120×13 (`benchmarks/bench_softplus_surface_fit.py`; legacy = the prior two-full-fits approach):

| surface | legacy | scipy ≥ 1.16 (`lm`) | scipy 1.11 floor (`trf`) |
|---|---|---|---|
| sharp asymmetric V | 14.8 s | 5.0 s (**2.9×**) | 7.3 s (2.0×) |
| wide curved V | 3.0 s | 0.46 s (**6.4×**) | 0.59 s (4.8×) |
| exactly in-family | 0.16 s | ~neutral | ~neutral |

Accuracy unchanged: `bench_softplus_surface_minima.py` still recovers the in-family `b∝1/T` surface to machine precision, and every `test_softplus_surface.py` assertion passes. New tests cover a robust loss and the `method` override. Verified green on **both** paths — current scipy (548 passed, `lm`) and the minimum-deps floor scipy 1.11.2 (`trf`) — including the noisy-convexity hypothesis test across many seeds.


landau/interpolate/softplus.py
```diff
@@ -359,14 +397,32 @@ class SoftplusSurface2DInterpolator(SurfaceInterpolator):
...
+    many temperatures are sampled (hundreds in practice).
+
+    **Solver.**  By default the coupled solve picks its least-squares method by
+    scipy version (see :data:`_LM_HAS_INTERNAL_SCALING`); set :attr:`method` to
+    ``'lm'`` or ``'trf'`` to override.  scipy 1.16 switched the default variable
+    scaling of ``least_squares(method='lm')`` to MINPACK's internal scaling
+    (``mode=1``), matching what plain ``leastsq`` always did -- see the `scipy
+    1.16.0 release notes
+    <https://docs.scipy.org/doc/scipy/release/1.16.0-notes.html>`_, gh-22790 and
+    the bug it fixed, gh-19459.  Before 1.16, ``lm`` silently disabled that
```
pmrv:
```suggestion
    (``mode=1``), matching what plain ``leastsq`` always did -- see gh-22790 and
    the bug it fixed, gh-19459 over on gh/scipy.  Before 1.16, ``lm`` silently disabled that
```

landau/interpolate/softplus.py
```diff
@@ -499,29 +563,79 @@ def fit(self, T, c, f) -> SoftplusFittedSurface:
...
+            a0, b0, c0, off0 = _fit_slice(cn[m], f[m], n, seed_nfev)
+            seeds.append(self._const_init(a0, b0, c0, off0))
+
+        method = self.method
+        if method is None:
+            method = "lm" if (_LM_HAS_INTERNAL_SCALING and self.loss == "linear") else "trf"
+        if method == "lm":
+            solver_kw = dict(method="lm")  # __post_init__ guarantees loss == "linear"
+        else:
+            solver_kw = dict(method="trf", loss=self.loss, f_scale=self.f_scale, x_scale="jac")
```
pmrv: abstract into method

landau/interpolate/softplus.py
```diff
@@ -597,13 +612,7 @@ def fit(self, T, c, f) -> SoftplusFittedSurface:
...
             seeds.append(self._const_init(a0, b0, c0, off0))
 
-        method = self.method
-        if method is None:
-            method = "lm" if (_LM_HAS_INTERNAL_SCALING and self.loss == "linear") else "trf"
-        if method == "lm":
-            solver_kw = dict(method="lm")  # __post_init__ guarantees loss == "linear"
-        else:
-            solver_kw = dict(method="trf", loss=self.loss, f_scale=self.f_scale, x_scale="jac")
+        solver_kw = self._solver_kwargs()
```
pmrv: call directly where you call the solver, don't litter local vars

pmrv (commented):
lgtm, merge

PR #304, opened 2026-06-30
fix(refine): phase-set-keyed, run-local triple-point dedup

### LINKED ISSUE #210: Move DelaunayTripleRefiner dedup state out of solve() into run()

pmrv:
Closes #210. Investigates the report that the refined-point deduplication is over-zealous.

## Findings

`DelaunayTripleRefiner` deduplicated located triple points inside `solve()` against a mutable instance attribute `self._found`, using each candidate simplex's full `(T, mu)` extent as a rectangular tolerance. Two over-zealous failure modes, one of them demonstrable:

1. **State never reset across `run()` calls.** `_found` is set in `__init__` and never cleared, while the refiner uses base `Refiner.run()`. Reusing one instance — or just calling `run()` twice — makes the second pass drop real triple points that coincide with the first pass's results. Reproduction: the same instance run twice on the same diagram emits the triple point the first time and **nothing** the second.
2. **Dedup ignored phase identity.** The comparison was purely on `(T, mu)`. Two genuinely distinct invariants (e.g. an A-B-C and a B-C-D triple point near a quadruple point) sitting within one simplex extent of each other collapsed into one, even though they are different invariant reactions.

For line phases the *grid-driven* form of (2) is geometrically self-limiting (bracketing both invariants with correct-triple simplices needs resolution finer than their separation, which shrinks the tolerance below it), but the phase-blind merge is still wrong whenever the located points land within a candidate's extent — confirmed at the `run()` level with two distinct invariants ~5 K / ~6 meV apart.

This also reconciles the source with the tracker: PR #256 (plain `run()` move) and PR #259 (barycentric "owning simplex" attribution — premise was geometrically false, since a real invariant lands on the shared corner *outside* every simplex) were both **closed unmerged**, so the live code still carried the original in-`solve()` dedup. CLAUDE.md claimed #256 had merged; corrected here.

## Change

- `solve()` is now a pure function of `(cand, phases)` — it returns the located point and holds no state.
- The dedup moves into a thin `run()` override, **keyed on the coexisting-phase set** and **run-local**. Two simplices that bracket one invariant share its phase set and collapse to a single point; distinct invariants involve different phases and both survive.
- Dominated / negative-`T` points are filtered **before** the dedup, so a spurious minimum can no longer reserve a slot that hides a real neighbour (previously such points were appended to `_found` first).

## Tests

`tests/unit/test_refine.py`:
- `test_delaunay_triple_refiner_keeps_distinct_phase_sets` — two invariants (A-B-C, B-C-D) within one simplex extent of each other both emit at their analytic locations (fails on the old phase-blind dedup, which yields one point).
- `test_delaunay_triple_refiner_solve_is_pure` — repeated `solve()` on one candidate returns the same point instead of going silently empty.
- `test_delaunay_triple_refiner_run_is_repeatable` — the same instance applied twice to one diagram yields identical output.

Existing `test_delaunay_triple_refiner_deduplicates` and `test_boundary_id_delaunay_triple_rows_share_id` stay green. `pytest tests/unit/test_refine.py tests/unit/test_calculate.py tests/unit/plot/test_triplepoint.py tests/integration/test_border_coverage.py tests/regression tests/unit/test_resample.py` → all pass.

---
_Generated by [Claude Code](https://claude.ai/code/session_01XxYCU5LR6NpezjvTbLsg99)_


pmrv:
repeats previously rejected work

PR #308, opened 2026-06-30
feat(refine): add dc_max concentration-drift floor to CC tracers

pmrv:
## Problem

`ClausiusClapeyronRefiner` samples the solid/liquid line much more finely than an intermetallic boundary in the same diagram (the attached report figure).

The step size comes from `_CCBase._dT_adapt`, which bounds the **mu** drift per step: `half_width / |dmu/dT|`, clamped to `[dT_min, dT_max]`. That keeps the corrector bracket valid, but it ties the T-sampling density to the coexistence slope. By Clausius-Clapeyron, `dmu/dT = -Δs/Δc`, so a boundary between phases of near-equal entropy is flat in mu (`dmu/dT → 0`); the step then always saturates at `dT_max` and the boundary is sampled coarsely — even where its concentration sweeps a lot, which is exactly what the c-T diagram plots.

Instrumenting a reproduction (terminal + intermetallic + ideal-solution liquid, 40×60 grid) confirms it: ~50% of CC steps have `|dmu/dT| < 1e-6` and step at the full `dT_max`, while the sloped liquid line takes small steps and is dense.

## Change

Add a `dc_max` concentration-drift cap shared by both `_CCBase` tracers (`ClausiusClapeyronRefiner`, `MiscibilityGapRefiner`), and tighten the step bounds:

- **Defaults `dT_max = 5 K`, `dc_max = 0.01` (1% drift per step)** (were 50 K / no cap).
- The step shrinks so the plotted concentration moves at most `dc_max` per step, using the drift observed over the previous step and primed from the seed so no coarse bootstrap step slips through.
- The cap is applied **after** the `[dT_min, dT_max]` clamp, so `dT_min` yields when it would otherwise block the drift target. The bootstrap `dT_min` kick is re-solved once at the concentration-capped step for the same reason, so the 1% bound holds from the first step. A tiny absolute floor keeps the walk progressing.
- A straight constant-c boundary keeps `dc/dT ≈ 0`, so the cap never engages and nothing is over-sampled. Existing diagrams are unchanged in shape; the full suite (and the testplot render) pass.

`_emitted_concentrations` reads the plotted concentrations generically off either emitted type (`RefinedPoint` → one branch per coexisting phase, `RefinedMiscibilityGap` → its two stored branches), so the base trace loop needs no per-subclass hook.

## Tests

`tests/unit/test_refine.py` uses a toy phase whose composition drifts with T at fixed mu, so the located boundary is exactly flat in mu (`|mu| < 1e-9`) while c sweeps — isolating the regime the cap targets:

- `test_cc_refiner_dc_max_bounds_concentration_drift` — on a short T window where the cap binds below `dT_min`, every consecutive step bounds the per-step concentration jump to `dc_max`, `dT_min` yields (sub-1 K steps), and the walk isn't truncated.
- `test_cc_refiner_dc_max_densifies_curved_boundary` — tightening `dc_max` adds samples the mu-drift heuristic alone leaves coarse.
- `test_cc_refiner_dc_max_noop_on_constant_c_boundary` — a constant-c boundary gets the same point count cap-on vs cap-off.

The straddle-dedup test's point-count bound is rescaled to the finer default density.

`benchmarks/cc_sampling_density.py` reproduces the old-vs-new-defaults contrast on a flat-mu, c-sweeping boundary: 9 points / 10% worst-step drift → 63 points / exactly 1%.

Full suite: 560 passed, 1 xfailed.

### DIFF
--- a/landau/refine.py
+++ b/landau/refine.py
@@ -658,13 +658,24 @@ class _CCBase(Refiner):
     max_steps : int
         Hard cap on steps per trace direction, just in case the
         adaptive logic somehow fails to terminate.
+    dc_max : float
+        Maximum concentration drift allowed per step. ``_dT_adapt``
+        bounds the *mu* drift, which keeps the corrector bracket valid
+        but leaves the T-sampling density proportional to ``dmu/dT``: a
+        boundary that is nearly flat in mu (``dmu/dT -> 0``, so the step
+        saturates at ``dT_max``) but curves in c gets sampled coarsely.
+        Capping the per-step concentration drift gives the trace a
+        density floor tied to the curve that is actually plotted; a
+        straight constant-c boundary keeps ``dc/dT ~ 0`` and the cap
+        never engages, so nothing is over-sampled.
     """
 
     def __init__(self, dT_max: float = 50.0, dT_min: float = 1.0,
-                 max_steps: int = 500):
+                 max_steps: int = 500, dc_max: float = 0.02):
         self.dT_max = dT_max
         self.dT_min = dT_min
         self.max_steps = max_steps
+        self.dc_max = dc_max
 
     # -- subclass hooks -----------------------------------------------------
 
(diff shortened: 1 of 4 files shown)


landau/refine.py
```diff
@@ -658,13 +658,24 @@ class _CCBase(Refiner):
...
+        saturates at ``dT_max``) but curves in c gets sampled coarsely.
+        Capping the per-step concentration drift gives the trace a
+        density floor tied to the curve that is actually plotted; a
+        straight constant-c boundary keeps ``dc/dT ~ 0`` and the cap
+        never engages, so nothing is over-sampled.
     """
 
     def __init__(self, dT_max: float = 50.0, dT_min: float = 1.0,
-                 max_steps: int = 500):
+                 max_steps: int = 500, dc_max: float = 0.02):
```
pmrv: let's update both max steps, I want at most 5K and 0.01% drift per refinement

PR #312, opened 2026-06-30
feat(refine): add dc_min concentration-drift floor; kwarg-only CC ctors

pmrv:
## Change

Follow-up to #308, which added a `dc_max` concentration-drift cap (a density *floor*) to the `_CCBase` tracers. This adds the complement, `dc_min`, plus a small constructor cleanup.

- **`dc_min` density ceiling.** Where a coexistence boundary is steep in mu — so `_dT_adapt` shrinks the step toward `dT_min` — but nearly flat in c, `dc_max` never engages and the curve is over-sampled with redundant points. `dc_min` grows the step so the plotted concentration moves at least `dc_min` per step, capped at `dT_max`. Shared by `ClausiusClapeyronRefiner` and `MiscibilityGapRefiner` via `_CCBase`, reading the plotted concentrations through the existing `_emitted_concentrations` hook.
- **`max` bounds take precedence over `min` bounds.** The `dc_min` floor is applied before the `dc_max` cap in the trace loop, so `dc_max` always wins: a `dc_min` set at or above `dc_max` yields and the per-step drift never exceeds `dc_max`. `dT_max` likewise bounds the floor.
- **Disabled by default (`0.0`)**, so it is a no-op unless set and existing diagram shapes are unchanged.
- **Keyword-only constructors, reordered args.** `_CCBase.__init__` and `MiscibilityGapRefiner.__init__` are now keyword-only (`*`), with the dc caps grouped and the `max_steps` hard cap moved last: `dT_max, dT_min, dc_max, dc_min, max_steps`. Every existing caller (tests, `benchmarks/cc_sampling_density.py`) already passes by keyword, so nothing breaks.

## Tests

`tests/unit/test_refine.py` adds a toy `_SteepMuFlatCPhase` whose boundary mu-slope (set by `semigrand_potential`) and plotted concentration (set independently by `concentration`) are decoupled, isolating the steep-in-mu / flat-in-c regime the floor targets:

- `test_cc_refiner_dc_min_floors_concentration_drift` — each steady step's plotted drift is pinned to exactly `dc_min` (`dT = dc_min / cslope`), nothing drifts more, asserted at `atol=1e-9`.
- `test_cc_refiner_dc_min_thins_oversampled_boundary` — turning the floor on coarsens an over-sampled boundary to fewer points.
- `test_cc_refiner_dc_min_noop_when_drift_already_met` — a boundary whose bare step already drifts past `dc_min` gets the identical point count whether the floor is set or off.
- `test_cc_refiner_dc_max_takes_precedence_over_dc_min` — with `dc_min` set 5× larger than `dc_max`, the per-step drift is still capped at `dc_max`.
- `test_cc_refiner_dT_max_bounds_dc_min_floor` — the floor never oversteps `dT_max` when the drift target would call for a larger step.

Full suite: 567 passed, 1 xfailed.

### DIFF
--- a/landau/refine.py
+++ b/landau/refine.py
@@ -673,9 +673,12 @@ class _CCBase(Refiner):
         steep in mu (``_dT_adapt`` shrinks the step toward ``dT_min``) but
         nearly flat in c, where ``dc_max`` never engages and the curve
         gets over-sampled. The step grows so the plotted concentration
-        moves at least ``dc_min``, but never past ``dT_max``. Disabled by
-        default (``0.0``); set it below ``dc_max`` to thin redundant
-        points without losing resolution.
+        moves at least ``dc_min``, but never past ``dT_max``. The ``max``
+        bounds take precedence over the ``min`` bounds: the floor is
+        applied before the ``dc_max`` cap, so a ``dc_min`` set at or above
+        ``dc_max`` yields to it and the per-step drift never exceeds
+        ``dc_max``. Disabled by default (``0.0``); set it below ``dc_max``
+        to thin redundant points without losing resolution.
     max_steps : int
         Hard cap on steps per trace direction, just in case the
         adaptive logic somehow fails to terminate.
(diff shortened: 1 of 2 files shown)


pmrv (commented):
both max parameters should have precedence to the min parameters, ensure that in the tests

pmrv:
Done in 950f799. The `dc_min` floor is now applied before the `dc_max` cap in the trace loop, so `dc_max` wins — a `dc_min` set at or above `dc_max` yields and the per-step drift never exceeds `dc_max`; `dT_max` likewise bounds the floor. Pinned by `test_cc_refiner_dc_max_takes_precedence_over_dc_min` (drift capped at `dc_max` with `dc_min` set 5× larger) and `test_cc_refiner_dT_max_bounds_dc_min_floor`.

---

pmrv:
the precedence should also apply cross wise, ie. a high dc_min shouldn't be able to force dT larger than dT_max.

pmrv:
That cross-wise case is already enforced. The floor clamps its target to `dT_max`:

```python
if dc_dT > 0 and self.dc_min > 0:
    dT_adapt = min(max(dT_adapt, self.dc_min / dc_dT), self.dT_max)
```

so `dc_min` can never drive `dT` past `dT_max`. `test_cc_refiner_dT_max_bounds_dc_min_floor` (in 950f799) pins it: `dc_min=0.05` calls for a 50 K step but the trace is held to `dT_max=5 K`. The symmetric pair — `dc_max` allowed below `dT_min` (`dT_min` yields) — is the existing #308 behavior. Let me know if you meant a different interaction.

---

PR #324, opened 2026-07-05
feat(phases): CompoundEnergyPhase (binary CEF) + Au-Cu reproduction

pmrv:
A minimal, general binary Compound Energy Formalism phase (`landau/phases/compoundenergy.py`), a real `Phase`, reproducing the Au-Cu system (Sundman, Fries &amp; Oates, *Calphad* 22 (1998) 335). Supersedes the earlier `BinaryCompoundEnergyPhase` `free_energy`-only sketches (issue #138 branches). Stacked on #345 (the `_dominated` refiner fix, which surfaced through this diagram); the speed work and the clamping/single-assessment refinements are stacked on top in #326.

## Model

`G(y,T) = Σ_endmembers °G_c(T)·∏_s p_s + kB·T·Σ_s a_s (y_s ln y_s + (1-y_s) ln(1-y_s)) + excess(y,T)`, with `y_s` the B site fraction on sublattice `s`, `a_s = site_multiplicities` (sum to 1). Fields: `site_multiplicities`, `endmember_energies` (all `2**S` config-tuples → `G(T)`), `excess(y,T)` (broadcasts over leading axes), and the seed-control pair `orderings`/`include_disordered_seed`. All energies per atom (eV); convert CALPHAD J/mol by `/96485.33`. Hashes/compares by dataclass fields like the other phases.

## Solver — one minimisation, straight from site fractions

The semi-grand potential is the Legendre transform `phi = min_y [G(y,T) - dmu·c(y)]` over `y ∈ [0,1]^S`. Its stationarity condition is the self-consistent Bragg-Williams (BGBW) fixed point `y_s = sigmoid((dmu - a_s^-1 ∂(G_ref+EG)/∂y_s)/kB T)` (the ideal term is linear in `logit(y_s)`), iterated damped and **vectorised over the whole dmu array** (one solve per T) from a disordered start plus every ordered corner, keeping the lowest `phi` per dmu. `G_ref` gradient analytic; `excess` gradient central-difference. `semigrand_potential`/`concentration` share one cached solve. First-order order/disorder transitions surface as `c(dmu)` jumps. (#326 batches this over `(T, dmu)` and replaces the damped iteration with a logit-space Steffensen/Aitken solve.)

## Competing ordered/disordered phases (no degeneracy)

[...]

PR #326, opened 2026-07-05
perf(phases): speed up CompoundEnergyPhase solve for full diagrams

pmrv:
Stacked on #324. Makes the `CompoundEnergyPhase` solver fast and its *refined* Au-Cu diagram correct, and simplifies the benchmark to the paper-faithful Assessment I. A full Assessment I (liquid + 4 fcc phases, 300–1400 K) refines with the triple-point and Clausius-Clapeyron refiners in **~20 s cold**; the diagram is unchanged across the solver rewrites (identical stable-point counts, both eutectoids).

Net changes over #324:

## Solver

- **Batched grid solve** — one vectorised SCF pass over the whole `(T, dmu)` mesh (`_solve_batch`) instead of a Python loop over unique `T`; `_free_energy` / `_grad_non_ideal` / `_scf` take a per-element `T`. Bit-identical to the per-`T` corner-seeded loop (`max|Δc| = 0`, `max|Δφ| ~ 1e-17`).
- **Steffensen's method with Aitken-`Δ²` in logit space** — replaces the damped fixed-point iteration. Each cycle applies the map twice and extrapolates `u = logit(y)` element-wise. The plain map converges only linearly and, near the fcc spinodal (the disordered branch between two ordered domes), oscillates period-2; extrapolating in `u` (unbounded) rides through the oscillation and accelerates the linear rate several-fold, while `sigmoid` maps it straight back into `(0, 1)` so `y` can never leave range — the failure mode of a bare Aitken step, which overshoots to `nan`. A pinned phase confines `u` to its ordering quadrant's logit box. **138450 → 90732 map evaluations** on the full Assessment I refine.
- **Warm-start caching** — scalar-`T` refinement probes seed the SCF from the nearest cached grid slice (`_warm`); a pinned phase solves **cold from its ordered corner** (deterministic). Solve cache `_find_phi_c_cached` maxsize 512 → 50000; one-call `excess` gradient (stack the `2·S` perturbations); active-column masking.

## Distinct ordered phases by a site-fraction clamp, not a `+inf` absence flag

#324 keeps the pinned ordered phases distinct by reporting `phi = +inf` wherever their long-range order falls below a threshold. That discontinuity was the source of a class of refinement artifacts: a triple point placed an ordered phase at `c = 0`, a Clausius-Clapeyron trace overstepped the eutectoid (fcc drawn coexisting with each intermetallic below the invariant), and the ordered domes' metastable continuations aliased against the disordered branch.

Replace it with a **site-fraction clamp**: each pinned phase's B-rich sublattices are held above 0.5 and its A-rich ones below (by `_y_margin`), so it stays in its ordering's quadrant. Its `phi` is finite and continuous everywhere and it simply loses the argmin outside its dome — no discontinuity for the refiners to trip on. `_order_tol` / the absence flag are gone.

## Benchmark — Assessment I only

- The paper's **Assessment II is a separate substitutional refit that drops ordering** (its own liquid + fcc, eq 10, no ordered phases). The benchmark had grafted Assessment I's ordered CEF phases onto Assessment II's substitutional fcc; once the absence flag was gone that mix is exposed as inconsistent — the two disordered-fcc descriptions differ by ~10 meV, so the ordered phases have no consistent disordered cap and balloon up to ~1200 K. **Dropped**; the benchmark now draws the single, paper-faithful Assessment I.
- `DelaunayTripleRefiner` + `ClausiusClapeyronRefiner` (loosened `dc_max = 0.05`) locate the invariants and trace the two-phase boundaries; default `poly_method` resolves to `segment-fasttsp` (smooth ordered-dome caps instead of the segment stitcher's spiky rings).

## Tests

147 `test_phases.py` + 52 `test_refine.py` pass. The pinned-phase tests assert the clamping behaviour (finite-but-metastable outside its dome; L1_0 and AuCu3 do not tie at `c ≈ 0.5`; the disordered phase finite everywhere).

Follow-up parked in #344 (modelling fcc+ordered as one partitioning phase with `MiscibilityGapRefiner`; the phase is equal-cost, the refiner is the blocker).


pmrv:
try to vectorize in T as well.

pmrv:
Done in 6fff4dc. The grid solve now batches over the whole `(T, dmu)` mesh in one SCF pass (`_solve_batch`) instead of looping over unique `T`; `_free_energy`/`_grad_non_ideal`/`_scf` take a per-element `T`. Bit-identical to the old per-`T` corner-seeded loop (`max|Δc| = 0`).

Profiling flagged that this only helps the grid (now ~1–2 s). The wall-clock is the boundary tracing: each refinement probe is a scalar-`T` solve through `brentq`, and vectorising the grid doesn't reach it. A single assessment is ~35 s cold, not the ~23 s the previous PR body claimed — that number was from a warm cross-instance cache. Corrected the PR body. Cutting refinement further would mean batching the refiner's probes.

---

PR #334, opened 2026-07-07
Vectorized coarse-then-refine CC trace (parked, with timing analysis)

pmrv:
Parked record of a vectorized rewrite of the `ClausiusClapeyronRefiner` trace. It is correct and fully tested but does **not** speed up the refiner on the phases currently in the library, and regresses expensive ones. Kept because the mechanism is sound and the win lands once a phase whose `semigrand_potential` batches over an array of distinct `T` (with non-trivial per-element cost) exists. Not proposed for merge as-is.

Rebased onto main past #335 (numpy-backed `_Simplex`), #340 (CC trace early abort) and #341 (shared Delaunay tessellation); every number below re-measured at commit c21ac9e against main 26a295b on the same machine.

## What changed

`ClausiusClapeyronRefiner.solve` (only — `MiscibilityGapRefiner` keeps the shared `_CCBase._trace`):

1. **Coarse walk** — a handful of big `dT` steps in both T directions to find the coexistence line's extent (`_coarse_walk`, scalar corrector). The walk aborts once a walked point is dominated and `solve()` emits only globally stable points — the invariant #340 pinned for the sequential tracer (`test_cc_refiner_trace_aborts_in_dominated_region`); adapted in 2948738.
2. **Recursive densification** — repeatedly halve every interval still coarser than `dT_max`/`dc_max` and locate all of one level's midpoints in a single vectorized pass over `semigrand_potential` (`_densify` → `_bisect_vectorized`).

New module helpers `_eval_over_arrays` (array-`T` with an elementwise fallback for phases like `RegularSolution` that key a fit on scalar `T`) and `_bisect_vectorized` (per-element bracketed bisection, NaN where no sign change).

`dc_max` drives subdivision (overriding `dT_min`, as its cap did before). The coarse-first walk caps sampling at the `dT_max` scale in flat-c regions — which is what `dc_min` did for the sequential tracer — so `dT_min` and `dc_min` are **inert for this refiner** (still accepted, still live for `MiscibilityGapRefiner`).

## Timing analysis

All numbers from `benchmarks/bench_cc_vectorized.py` (commit c21ac9e), 5-run best, `refine_phase_diagram` with only the CC refiner. The end-to-end table and the eval attribution are the script's measurements [0] and [2] run once on `main` (26a295b) and once on this branch.

**End-to-end wall-clock**

| system | old (main) | new | output points old → new |
|---|---|---|---|
| `IdealSolution` solid/liquid (cheap `phi`) | 90 ms | 92 ms | 211 → 514 |
| `FastInterpolatingPhase` pair (expensive `phi`) | 29 ms | 855 ms | 10 → 10 |

Cheap phase: a wash on wall-clock, at 2.4× the emitted boundary density (the densification subdivides to the `dc_max` target rather than stepping at the adaptive `dT`). Expensive phase: ~30× slower — main's baseline has dropped from 255 ms to 29 ms since the first measurement here (#335 removed the per-simplex `df.iloc` cost, #340 stopped tracing metastable tails), so the scalar coarse-walk overhead stands out far more starkly than the ~4× first reported.

**Why — three diagnostics from the benchmark:**

1. *`propose()` is no longer a bottleneck anywhere.* On the ideal grid (525 points, 960 simplices) `propose()` is 2.7 ms of a 74 ms `run()` (4%) after #335 — at the first measurement it was 198 ms of 272 ms (73%). The trace's `semigrand_potential` calls now dominate for cheap phases too, so vectorizing them is aimed at the right number; it just doesn't win on these phases.

2. *The extra cost is the scalar coarse walk — which costs more evals than the whole sequential tracer it replaces, despite walking fewer points.* Same 10-point case, `_solve_fixed_T` evals per stage (measurement [2] on both checkouts):

   | | seed | extent walk | densify | total |
   |---|---|---|---|---|
   | main (`_trace`) | 80 | 80 | — | **160** |
   | this branch (`_coarse_walk`) | 80 | 944 | 20 | **1044** |

   The walk's T-step is genuinely coarse (span/8 = 200 K here), but on a candidate whose line has no extent each extension attempt retries the corrector with up to 12 bracket doublings before ending the walk — 10 candidates × 2 directions × 12 attempts ≈ the 944 evals — where the sequential tracer makes one corrector attempt per direction, fails to bracket, and stops (80). Coarser per point, ~12× denser per eval on this case.

3. *The densification does batch where the phase vectorizes over T.* On `IdealSolution` (broadcasts over `T`), the densification issues `semigrand_potential` calls of up to **128 elements** (880 of 3354 calls are multi-element; the total grew from 2832 by the ~500 scalar `_dominated` stability checks added for the #340 semantics). The vectorization the design targets does happen — it just isn't the dominant cost today.

## Dedup / eval-count anomaly

Investigated the "6× more evals, same 10 output points" flagged in review; re-verified unchanged after the rebase. The `FastInterpolatingPhase` test case is a **degenerate near-isothermal feature** at `T=725`: of 40 two-phase simplices, 10 solve and each emits a **single** point (span `[725,725]`); the other 30 straddle-skip. A one-point trace's bbox is too small to make the 10 co-located candidates straddle-skip each other, so all 10 run their coarse walk with 12× bracket expansion trying to extend a line that has no T-extent → the 944 walk evals above (the other 80 are the seed root-finds, identical on main). So: not a batching failure, but (a) coarse-walk bracket expansion is wasteful on off-line/degenerate candidates, and (b) the straddle dedup does not collapse co-located candidates on a near-isothermal feature. (b) predates this branch; (a) is this branch's walk amplifying a candidate set main's tracer bails out of cheaply.

## Known CI failure (expected on this branch)

One border-coverage integration case fails per CI run, and **which one varies between runs**. The latest-deps build jobs reproducibly fail `segments-c-T` (a stable border point 2.4% outside its phase's c-T polygon, threshold 2%; identical value on 3.11/3.12/3.13 and locally). The minimum-deps job, across two runs of test-identical code, failed once on `segments-c-T` (same 2.4%) and once on `tsp-mu-T` (5.4%, with `segments-c-T` passing); the pre-rebase branch failed `tsp-mu-T` at 3.9%. The TSP-based poly methods carry unseeded solver randomness, so the marginal point sets this branch's changed border sampling produces land either side of the 2% threshold depending on the run. All of these cases pass on current main (`Segments × μ-T` remains the only pre-existing xfail). Left unfixed by design: resolving it means cleaning up the coarse-walk sampling, i.e. reviving this parked work. It is the concrete blocker to address before this branch is picked back up.

## Tests

`tests/unit/test_refine.py` (58 pass, including main's #340 pin `test_cc_refiner_trace_aborts_in_dominated_region`, which 2948738 adapts the coarse walk to): direct tests for `_bisect_vectorized` (batch roots, bracket expansion, NaN on no sign change) and `_eval_over_arrays` (vectorized + scalar-only fallback); a densification-accuracy test (every refined midpoint on `phi_solid == phi_liquid` to `<1e-7`, and `dc_max` controls density); the `dc_min` block pins the coarse-first behavior (flat-c boundary capped at `dT_max` with `dc_min` off; `dc_min` inert; `dc_max` still caps drift). Full `unit`/`regression` suites green (633 pass); `integration` green (8 pass, 1 xfail) except the run-varying border-coverage case noted above.

## Where this goes

- #335, #340 and #341 have landed on main and are in this branch's baseline now. The sequential tracer they left behind is ~9× faster on the expensive-phase benchmark than when this branch was first measured, which raises the bar for reviving it.
- This branch is worth resuming if/when a phase with an internally-vectorized-over-`T` `semigrand_potential` lands; the coarse-walk waste (item 2 / anomaly-a / the border-coverage failure) should be trimmed first — bounding the bracket expansion (or reusing main's single-attempt bail-out) recovers most of the 944-vs-80 gap.

### DIFF
--- a/benchmarks/bench_cc_vectorized.py
+++ b/benchmarks/bench_cc_vectorized.py
@@ -1,0 +1,147 @@
+"""Where the ``ClausiusClapeyronRefiner`` trace spends its time.
+
+This driver backs the timing analysis on the vectorized coarse-then-refine CC
+trace (the ``_bisect_vectorized`` densification). It is a *diagnostic* record,
+not a win: on the phases currently in the library the vectorization does not
+speed the refiner up, and the numbers here say why. It becomes a win only for a
+phase whose ``semigrand_potential`` genuinely batches over an *array of
+distinct T* with non-trivial per-element cost.
+
+Three measurements, run from the repo root with
+``python benchmarks/bench_cc_vectorized.py``:
+
+1. **propose vs. trace (cheap phase).** On an ``IdealSolution`` solid/liquid
+   grid, ``propose`` (Delaunay + a ``df.iloc[simplex]`` per simplex) dominates;
+   the trace's ``semigrand_potential`` calls are a minority. Vectorizing the
+   trace cannot move a number the refiner is not spending time on.
+
+2. **eval attribution (expensive phase).** On a ``FastInterpolatingPhase`` pair
+   the expensive ``_solve_fixed_T`` calls are counted by refiner call site. They
+   land almost entirely in the *scalar* coarse walk / seed root-finding
+   (``<lambda>`` inside ``_refine_step``), not in the vectorized densification.
+
+3. **densification batching (vectorizing phase).** For ``IdealSolution`` —
+   which broadcasts ``semigrand_potential`` over ``T`` — the densification's
+   ``_bisect_vectorized`` calls carry many elements each, i.e. the batching the
+   design is built to exploit does happen; it just does not dominate runtime.
+"""
+from __future__ import annotations
+
+import time
+from collections import Counter
+import traceback
+
+import numpy as np
+from scipy.constants import Boltzmann, eV
+
+from landau.phases import LinePhase, IdealSolution, FastInterpolatingPhase
+from landau.interpolate import RedlichKister
+from landau.calculate import calc_phase_diagram, refine_phase_diagram
+from landau.refine import ClausiusClapeyronRefiner, _delaunay_simplices
+
+kB = Boltzmann / eV
+
+Ts = np.linspace(200, 1800, 25)
+mus = np.linspace(-0.3, 0.3, 21)
+
+
+def _best(fn, repeats=5):
+    best = np.inf
+    for _ in range(repeats):
+        t = time.perf_counter()
+        fn()
+        best = min(best, time.perf_counter() - t)
+    return best
+
+
+def ideal_system():
+    solid = IdealSolution(
+        "solid", LinePhase("A", 0, -2.0, 1.0 * kB), LinePhase("B", 1, -3.0, 1.5 * kB))
+    liquid = IdealSolution(
+        "liquid", LinePhase("A(l)", 0, -1.9, 2.5 * kB), LinePhase("B(l)", 1, -2.9, 2.2 * kB))
+    return [solid, liquid], {"solid": solid, "liquid": liquid}
+
+
+def fast_system():
+    def mk(name, e, s):
+        cs = [0, 0.25, 0.5, 0.75, 1.0]
+        lps = [LinePhase(f"{name}{i}", c, e + 0.2 * c * (1 - c), s) for i, c in enumerate(cs)]
+        return FastInterpolatingPhase(
+            name=name, phases=lps, add_entropy=True, interpolator=RedlichKister(2))
+    a, b = mk("alpha", -2.0, 1.0 * kB), mk("beta", -1.9, 2.6 * kB)
+    return [a, b], {"alpha": a, "beta": b}
+
+
+def propose_vs_trace():
+    phases, mapping = ideal_system()
+    df = calc_phase_diagram(phases, Ts=Ts, mu=mus, refine=False, keep_unstable=False)
+    r = ClausiusClapeyronRefiner()
+    n_simplices = len(list(_delaunay_simplices(df)))
+    t_propose = _best(lambda: list(r.propose(df)))
+    t_run = _best(lambda: r.run(df, mapping))
+    print("[1] propose vs. trace (IdealSolution, cheap phi)")
+    print(f"    grid={len(df)} points, {n_simplices} simplices")
+    print(f"    propose(): {t_propose * 1e3:6.1f} ms   "
+          f"full run(): {t_run * 1e3:6.1f} ms   "
+          f"propose share: {t_propose / t_run:4.0%}")
+
+
+def eval_attribution():
+    phases, mapping = fast_system()
+    df = calc_phase_diagram(phases, Ts=Ts, mu=mus, refine=False, keep_unstable=False)
+    site = Counter()
+    orig = FastInterpolatingPhase._solve_fixed_T
+
+    def counted(self, T, dmu):
+        last = "?"
+        for fr in traceback.extract_stack():
+            if "refine.py" in fr.filename:
+                last = fr.name
+        site[last] += 1
+        return orig(self, T, dmu)
+
+    FastInterpolatingPhase._solve_fixed_T = counted
+    try:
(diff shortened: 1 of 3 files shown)


pmrv:
Rebased onto 26a295b (#335/#340/#341) and re-verified every claim in the OP; body updated with numbers re-measured at f4a7fe2 on one machine.

- Rebase fallout: main's #340 now pins that CC `solve()` emits only globally stable points, which the coarse walk predated. 58da600 aborts the walk on domination and filters emitted points; the remaining `refine.py` delta vs main is exactly the two vectorized helpers plus the `ClausiusClapeyronRefiner` rewrite. The branch's `AGENTS.md` hunk turned out to be an accidental revert of #333's rows and is dropped by the rebase.
- Reproduced exactly: eval attribution (1044 = 1024 scalar coarse-walk + 20 vectorized, 10 output points), densification batching (max batch 128, 880 multi-element calls), and the `T=725` degenerate-feature anomaly (40 two-phase simplices, all solved traces single points at 725.3 K).
- Stale, corrected in the body: `propose()` share 73% → 4% (#335 landed); end-to-end old-vs-new is now 90 → 92 ms on `IdealSolution` (at 211 → 514 output points) and 29 → 855 ms on `FastInterpolatingPhase` — ~30× instead of the ~4× first reported, because main got ~9× faster on this benchmark since the branch was parked.
- The known CI failure moved: `tsp-mu-T` (3.9%) now passes; `segments-c-T` fails at 2.4% vs the 2% threshold instead. Both cases pass on current main.
- Suites at f4a7fe2: `test_refine.py` 58 pass, unit+regression 633 pass, integration 8 pass / 1 xfail / the `segments-c-T` case above.
- The end-to-end table's driver is now measurement [0] in `benchmarks/bench_cc_vectorized.py` rather than an unscripted checkout comparison.

---

pmrv:
History rewritten once more to drop a render PNG that had slipped into the adaptation commit: 58da600 → 2948738, f4a7fe2 → fc49749. Tracked content otherwise byte-identical; the OP cites the final SHAs.

---

pmrv:
> For expensive phases the extra cost is the scalar coarse walk, not a batching failure. For the 10-point case, _solve_fixed_T runs 1044 times: 1024 inside the scalar coarse-walk/seed root-finding and only 20 in the vectorized densification — unchanged by the rebase. The coarse walk's bracket-expansion is wasteful when a candidate is off the traceable line.

This sounds as if the 'coarse' walk is actually denser than the previous refiner, check that.

pmrv:
@claude [opus] (this is a new run, not for you, claude code web)

Investigate the Dedup / eval-count anomaly a new issue/PR.  Is this a pre existing failure or due to this PR?

pmrv:
Checked — you're right in eval terms. Same 10-point case, `_solve_fixed_T` evals attributed per stage, measurement [2] of the benchmark run on both checkouts (attribution fixed in c21ac9e so it reports per stage and runs on main too):

| | seed | extent walk | densify | total |
|---|---|---|---|---|
| main 26a295b (`_trace`) | 80 | 80 | — | 160 |
| this branch c21ac9e (`_coarse_walk`) | 80 | 944 | 20 | 1044 |

The walk is coarse in T-steps (span/8 = 200 K here, vs the sequential tracer's adaptive ≤5 K), but each failed extension attempt retries the corrector with up to 12 bracket doublings before ending the walk: 10 candidates × 2 directions × 12 attempts ≈ the 944 evals. Main's tracer makes one corrector attempt per direction, fails to bracket, and stops — 80 evals. So per walked point it is coarser; per eval it is ~12× denser than the whole previous trace on this degenerate case, and that retry loop, not point density, is where the regression lives. Seed cost is identical on both sides.

Corrected diagnostic 2 and the anomaly paragraph in the OP accordingly (the old text split the 1024 lambda-attributed evals wrongly and compared only branch-internal call sites, never against main).

---

PR #335, opened 2026-07-07
perf(refine): numpy-backed _Simplex instead of per-simplex df.iloc

pmrv:
The bottleneck surfaced while profiling the CC refiner (#334): every Delaunay refiner walks the `(mu, T)` tessellation through `_delaunay_simplices`, which materialized one `df.iloc[simplex]` DataFrame per simplex — ~960 three-row frames on a 25×21 grid. Building those frames dominated `propose()` and hence `run()`; a full `default_refiners` refine spent most of its wall-clock there (each of the three refiners rebuilds the tessellation).

## Change

`_delaunay_simplices` now yields a numpy-backed `_Simplex` view of the four columns the refiners actually read — `T`, `mu`, `phase`, `c` — as length-3 arrays, instead of a DataFrame. Consumers (`_phase_centroids_xy`, `_simplex_containment`, `_simplex_brackets`, and the `propose`/`solve` sites in all four refiners) index numpy directly. `_Simplex` is `eq=False` to preserve the identity semantics `DelaunayTripleRefiner.solve`'s `owner is not tr` relies on, and to sidestep the ambiguous-truth-value error a field-wise array `__eq__` would raise.

No change to the located transitions — the arithmetic is identical, just off numpy arrays. Micro-benchmarking the extraction confirmed a DataFrame is inherently the wrong container here: even building small frames from a numpy dict was *slower* than the original `df.iloc`; plain arrays are ~40× faster.

## Numbers

25×21 grid (960 simplices), 5-run best, from `benchmarks/bench_delaunay_propose.py`:

| measurement | before | after | speedup |
|---|---|---|---|
| `_delaunay_simplices` (materialize) | 157 ms | 9.7 ms | 16× |
| `ClausiusClapeyronRefiner.propose` | 191 ms | 12.5 ms | 15× |
| `ClausiusClapeyronRefiner.run` | 258 ms | 81 ms | 3.2× |
| `default_refiners` full refine | 969 ms | 120 ms | 8× |

The full-refine 8× is because all three default refiners each build the tessellation, so the per-simplex DataFrame cost was paid three times.

## Tests

- `test_delaunay_simplices_yields_numpy_backed_vertices` pins the new view: length-3 arrays, every vertex reproduces a source row, correct phase count, `unique_phases()` in appearance order.
- The `_phase_centroids_xy` and `_simplex_containment` direct helper tests now build `_Simplex` (via a small `_mk_simplex` helper) instead of DataFrames.
- Full `unit` / `regression` / `integration` suites unchanged (including `test_border_coverage.py`, the "polygon doesn't amputate a phase region" smoke test) — behavior is identical.

---
_Generated by [Claude Code](https://claude.ai/code/session_01J283DeTTzhoSz9j1Apn7vP)_

### DIFF
--- a/landau/refine.py
+++ b/landau/refine.py
@@ -363,28 +363,64 @@ class ScanRefiner(Refiner):
 # -- Delaunay-based refiners --------------------------------------------------
 
 
+@dataclass(frozen=True, eq=False)
+class _Simplex:
+    """Numpy-backed view of one Delaunay simplex's three vertices.
+
+    Yielded by :func:`_delaunay_simplices` in place of a per-simplex
+    ``df.iloc[...]`` DataFrame. Materialising ~960 three-row frames per
+    tessellation was the dominant cost of every Delaunay refiner's
+    ``propose`` (a full ``run`` spends the majority of its time there);
+    slicing the four columns the refiners actually read — ``T``, ``mu``,
+    ``phase``, ``c`` — as length-3 numpy arrays instead is ~40x cheaper.
+
+    ``eq=False`` keeps identity semantics (``owner is not tr`` in
+    :meth:`DelaunayTripleRefiner.solve` relies on it) and sidesteps the
+    ambiguous-truth-value error a field-wise array ``__eq__`` would raise.
+    """
+
+    T: np.ndarray
+    mu: np.ndarray
+    phase: np.ndarray
+    c: np.ndarray
+
+    def unique_phases(self) -> np.ndarray:
+        """Distinct phase names in vertex-appearance order (like ``pd.unique``)."""
+        return pd.unique(self.phase)
+
+
 def _delaunay_simplices(df: pd.DataFrame):
-    """Yield (simplex_df, phase_count) for each simplex of the (mu, T) tess."""
+    """Yield ``(_Simplex, phase_count)`` for each simplex of the (mu, T) tess."""
     dela = Delaunay(df[["mu", "T"]])
-    phases_arr = df.phase.to_numpy()[dela.simplices]
+    T = df["T"].to_numpy()
+    mu = df["mu"].to_numpy()
+    phase = df["phase"].to_numpy()
+    c = df["c"].to_numpy() if "c" in df.columns else np.zeros(len(df))
+    phases_arr = phase[dela.simplices]
     counts = np.array([len(set(x)) for x in phases_arr])
-    for simplex, n in zip(dela.simplices, counts):
-        yield df.iloc[simplex], n
+    for s, n in zip(dela.simplices, counts):
+        yield _Simplex(T=T[s], mu=mu[s], phase=phase[s], c=c[s]), int(n)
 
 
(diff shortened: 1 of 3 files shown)


landau/refine.py
```diff
@@ -363,28 +363,64 @@ def _solve_pair(self, name1, name2, bracket, potential, point, phases) -> list[R
...
+@dataclass(frozen=True, eq=False)
+class _Simplex:
+    """Numpy-backed view of one Delaunay simplex's three vertices.
+
+    Yielded by :func:`_delaunay_simplices` in place of a per-simplex
+    ``df.iloc[...]`` DataFrame. Materialising ~960 three-row frames per
+    tessellation was the dominant cost of every Delaunay refiner's
+    ``propose`` (a full ``run`` spends the majority of its time there);
+    slicing the four columns the refiners actually read — ``T``, ``mu``,
+    ``phase``, ``c`` — as length-3 numpy arrays instead is ~40x cheaper.
```
pmrv: numbers here are historic and will outdate.  just state that this class is faster to instantiate than dataframes.

landau/refine.py
```diff
@@ -363,28 +363,63 @@ def _solve_pair(self, name1, name2, bracket, potential, point, phases) -> list[R
...
-    :class:`ClausiusClapeyronRefiner` a reliable seed bracket.
+    :class:`ClausiusClapeyronRefiner` a reliable seed bracket. The two
+    centroids come out ordered by phase name (matching the old
+    ``groupby("phase").mean()``).
     """
-    by_phase = simplex.groupby("phase")[["T", "mu"]].mean().to_numpy()
-    return tuple(by_phase[0]), tuple(by_phase[1])
+    def centroid(name):
+        m = simplex.phase == name
+        return float(simplex.T[m].mean()), float(simplex.mu[m].mean())
```
pmrv: method on _Simplex, again do not talk about old code.

landau/refine.py
```diff
@@ -363,28 +363,63 @@ def _solve_pair(self, name1, name2, bracket, potential, point, phases) -> list[R
...
+
+    n1, n2 = np.unique(simplex.phase)
```
pmrv: use _Simplex.unique_phases

landau/refine.py
```diff
@@ -408,14 +408,13 @@ def _phase_centroids_xy(simplex: _Simplex) -> tuple[TMuPoint, TMuPoint]:
     other twice) the line from one centroid to the other is guaranteed
     to cross the phase boundary inside the simplex, which gives
     :class:`ClausiusClapeyronRefiner` a reliable seed bracket. The two
-    centroids come out ordered by phase name (matching the old
-    ``groupby("phase").mean()``).
+    centroids follow :meth:`_Simplex.unique_phases` order.
     """
     def centroid(name):
         m = simplex.phase == name
         return float(simplex.T[m].mean()), float(simplex.mu[m].mean())
```
pmrv: this should be a method, too `_Simplex.centroids -> dict[str, TMuPoint]`.

pmrv (commented):
lgtm merge after CI.

PR #340, opened 2026-07-08
perf(refine): abort CC trace when the pair goes metastable

pmrv:
Confirmed the reported behavior: `_CCBase._trace` walked each two-phase coexistence line all the way to the sampled T bound, and domination was only filtered afterward in `run()` (`if pt.T < 0 or _dominated(pt, phases): continue`). So the metastable tail past a triple point was fully traced — one `_refine_step` root-find per step — and then discarded.

## Change

`_CCBase._trace` now stops as soon as an emitted point is `_dominated` (a phase outside the pair is more stable there): the pair's line has crossed a triple point and gone metastable, so there is nothing worth tracing further. `solve()` likewise skips a seed that projects just past a triple point into a metastable sliver. Both are shared by `ClausiusClapeyronRefiner` and `MiscibilityGapRefiner`.

A stable segment that re-emerges past such a gap is still recovered: it has its own two-phase simplex, and because the shortened trace's straddle bbox no longer overlaps it, that seed is no longer skipped and traces the segment itself.

## No output change

`run()` already dropped these points, so the final refined dataframe is identical — the change only removes wasted root-finds. Verified by the full `unit` / `regression` / `integration` suites, including `test_border_coverage.py` (the "polygon doesn't amputate a phase region" smoke test), which is unchanged.

## Numbers

`benchmarks/bench_cc_early_abort.py`, three-phase system (triple point at T=300), 20×20 grid, 5-run best:

| | before (main) | after |
|---|---|---|
| `run()` | 40.2 ms | 33.6 ms |
| rows | 210 | 210 (identical) |

Per coexistence pair, the fraction of the sampled T-range no longer traced:

```
A-B:  300.0-480.0 K   31% skipped
B-C:  300.1-480.0 K   31% skipped
A-C:  220.0-299.8 K   69% skipped
```

The saving scales with how much of each line is metastable — larger on diagrams with more triple points / longer metastable tails.

## Tests

`test_cc_refiner_trace_aborts_in_dominated_region`: on the three-phase fixture, every point `solve()` emits (across all pairs, before `run()`'s filter) is globally stable, and the A-B trace stays on its stable side (T > 300) instead of walking down to T_min = 220. Fails on `main` (the trace emits the dominated tail), passes here.

---
_Generated by [Claude Code](https://claude.ai/code/session_01J283DeTTzhoSz9j1Apn7vP)_

### DIFF
--- a/benchmarks/bench_cc_early_abort.py
+++ b/benchmarks/bench_cc_early_abort.py
@@ -1,0 +1,91 @@
+"""Work skipped by the ClausiusClapeyronRefiner early abort.
+
+``_CCBase._trace`` used to walk each two-phase coexistence line all the way to
+the sampled T bound and let ``run`` discard the points where a third phase is
+more stable (the metastable tail past a triple point). The trace now stops as
+soon as it enters a dominated region, so those root-finds are never done. The
+final ``run`` output is unchanged — only the wasted work is removed.
+
+This driver traces a three-phase system whose A-B line is stable only above its
+triple point, and reports the run wall-clock plus, per coexistence pair, the
+fraction of the sampled T range that is metastable (and so no longer traced).
+
+Run ``python benchmarks/bench_cc_early_abort.py`` from the repo root. Measured
+against ``main`` (pre-abort) on the default grid: run() 40.2 ms -> 33.6 ms,
+identical 210-row output; the A-B trace shrinks from 220-480 K to 300-480 K.
+"""
+from __future__ import annotations
+
+import time
+
+import numpy as np
+import pandas as pd
+
+from landau.phases import LinePhase
+from landau.refine import ClausiusClapeyronRefiner, _dominated
+
+
+def three_phase_system():
+    """Three line phases meeting at a single triple point (T=300, mu=0.2)."""
+    return {
+        "A": LinePhase(name="A", fixed_concentration=0.0, line_energy=-1.0, line_entropy=0.004),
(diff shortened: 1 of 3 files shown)


pmrv:
@testplots all plots

pmrv:
check on the minimum tests.

pmrv:
The `test-minimum-deps` failure is unrelated to this PR — it's `test_softplus_surface.py::test_fitted_slice_is_convex_even_on_noisy_data`, a scipy-floor fit artifact:

- The diff here only touches `landau/refine.py`, its test, and a benchmark; that test and its `CONVEX_ATOL = 1e-9` are unchanged from `main`.
- The falsifying example (`seed=1248882415, Tq=1105`) reproduced locally: on the scipy floor the fitted slice's min second-difference is `-3.7e-9` (fails `>= -1e-9`), but on current scipy (1.17) the same example gives `+1.8e-6` — convex. So the floor's `least_squares` under-converges the convexity by a few ×1e-9 on that random draw; Hypothesis happens to hit it.
- `build (3.11/3.12/3.13)` all pass; local `unit`/`regression`/`integration` (incl. `test_border_coverage`) pass.

I don't have permission to re-run the job (`403` on rerun-failed-jobs). A fresh Hypothesis draw will very likely pass on re-run. If you'd rather remove the flakiness, I can open a separate PR hardening the softplus test on the floor (e.g. a small `CONVEX_ATOL` bump or tighter fit convergence) — kept out of this one to stay single-purpose.

---

PR #343, opened 2026-07-09
perf(refine): merge the 2-D default refiners into one Delaunay2DRefiner

pmrv:
Alternative to #341 (which memoises `_delaunay_simplices`). Instead of caching, this merges the three default 2-D refiners into one, as suggested — a single `run` that tessellates once and drives the three constituents.

## Design

`Delaunay2DRefiner` tessellates the `(mu, T)` grid **once** and dispatches each simplex by phase count:

- `n = 3` → `DelaunayTripleRefiner` (triple points)
- `n = 2` → `ClausiusClapeyronRefiner` (inter-phase coexistence lines)
- `n = 1` → `MiscibilityGapRefiner` (intra-phase gaps)

`run` builds the tessellation and calls each constituent's `run_one(df, simplices, phases)` on that shared list, so each keeps its own locate-and-emit bookkeeping — the Clausius-Clapeyron straddle dedup, the triple-point owner attribution — and its own `refined` label. It composes the three existing refiners rather than reimplementing them, so their tested logic is reused unchanged.

To make that possible (per review — `propose`'s signature is left alone):
- Each Delaunay-based refiner factors its `propose` body into a `_candidates(df, simplices)` generator. `propose(df)` stays `propose(self, df)` and delegates to `_candidates(df, _delaunay_simplices(df))`.
- `Refiner.run_one(df, simplices, phases)` is factored out of `run` — the `propose → solve → emit` path run off a *given* tessellation (`_run_candidates(self._candidates(df, simplices), phases)`). `Refiner.run` and the `_CCBase` straddle-dedup version of `_run_candidates` are unchanged.
- `default_refiners` returns `[Delaunay2DRefiner()]` for a 2-D grid. Constituents are swappable via `Delaunay2DRefiner(triple=…, clausius=…, gap=…)`.

Standalone refiners and user-defined `Refiner` subclasses are unaffected — `propose(self, df)` is unchanged and `run` still calls `self.propose(df)`.

## No output change

The emitted rows — located transitions, `refined` labels, boundary ids — are identical to running the three refiners separately. Unit, regression, and integration suites (incl. `test_border_coverage`) pass unchanged.

## Numbers

`benchmarks/bench_combined_refiner.py` (added), `refine_phase_diagram` only, best of 6, asserting combined vs separate rows are identical:

| system | 3 separate | combined | |
|---|---|---|---|
| `2d_basics_mu` (100×100, cheap phi) | ~950 ms | ~500 ms | ~1.9× (tessellation-bound) |
| `2d_toy` (FastInterpolatingPhase) | ~1200 ms | ~1100 ms | ~6% (per-T-solve-bound) |

`basics_mu` is tessellation-bound so it nearly halves; `toy` is dominated by the CC trace's per-T solves so the tessellation is a smaller slice.

## vs. #341

Both remove the 3× tessellation. #341 is a smaller diff (a weakref last-call memo on `_delaunay_simplices`, no API change) but adds module-level cache state. This is a larger diff but stateless — the tessellation is shared by construction and there is one refiner in `default_refiners`. Pick whichever fits; I can close the other.

## Tests

`test_delaunay_2d_refiner_matches_separate_refiners` (row-for-row equality vs the three separate), `test_delaunay_2d_refiner_shares_one_tessellation` (one `_delaunay_simplices` call for a whole `run` and a whole `propose`), `test_default_refiners_2d_is_single_combined_refiner`, and a constituent-swap test.

### DIFF
--- a/benchmarks/bench_combined_refiner.py
+++ b/benchmarks/bench_combined_refiner.py
@@ -1,0 +1,96 @@
+"""Delaunay2DRefiner vs. the three separate default refiners.
+
+The 2-D default used to run DelaunayTripleRefiner, ClausiusClapeyronRefiner and
+MiscibilityGapRefiner separately, each tessellating the (mu, T) grid in its own
+propose. Delaunay2DRefiner merges them: one tessellation, one propose that
+yields typed candidates, one run that buckets them back per refiner. This
+driver times both on the testplots example systems and checks the emitted
+transition rows are identical.
+
+Run ``python benchmarks/bench_combined_refiner.py`` from the repo root.
+Reference (best of 6, refine_phase_diagram only, on a precomputed coarse frame):
+
+    system        3 separate   combined
+    2d_basics_mu      949 ms      498 ms   (~1.9x; tessellation-bound)
+    2d_toy           1205 ms     1128 ms   (~6%; per-T-solve-bound)
+"""
+from __future__ import annotations
+
+import time
+
+import numpy as np
+
+import landau.phases as ldp
+import landau.calculate as ldc
+import landau.interpolate as ldi
+from landau.refine import (
+    Delaunay2DRefiner, DelaunayTripleRefiner, ClausiusClapeyronRefiner,
+    MiscibilityGapRefiner,
+)
+
+SEPARATE = [DelaunayTripleRefiner(), ClausiusClapeyronRefiner(), MiscibilityGapRefiner()]
+COMBINED = [Delaunay2DRefiner()]
+
+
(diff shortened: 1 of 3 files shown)


pmrv (commented):
I don't like messing with propose's signature.  Instead factor out a run_one from the three individuals run() and call those in Delaunay2D's run.

pmrv (commented):
you took a hard left somehwere, close this and merge the competitor.

pmrv:
Closing in favor of #341 (merged), per review.

---

PR #345, opened 2026-07-09
fix(refine): _dominated drops non-coexisting refined transitions robustly

pmrv:
Two fixes to `_dominated` in `landau/refine.py` — the predicate every refiner's `run()` uses to drop a refined transition whose phases aren't the globally stable set at `(T, mu)`. Broken out of the CEF work (#324) since it's a general refiner-correctness change; #324/#326 stack on top.

## The bug

`_dominated` compared every phase to `next(iter(own))` — an arbitrary member of a **set** — and only checked phases *outside* the transition. So a three-phase Delaunay simplex could keep a phantom vertex whose phase is absent (`phi = +inf`) or otherwise not co-stable, and whether that vertex survived depended on set-iteration order — **nondeterministic**. In the Au-Cu CEF diagram this surfaced as a spurious triple point placing an ordered phase at `c = 0` (a thin spike to the axis) that appeared in some runs and not others.

## The fix

Recast the predicate as "are the own phases the most-stable set here":

- Drop when an own phase is **absent** (`phi = +inf`) — a phase that doesn't exist here can't be a coexisting vertex.
- Drop when a phase **outside** the transition sits below the *least-stable* own phase (`max(own_phi)`), i.e. the own set isn't the top-|own| most stable. Comparing to the maximum (not an arbitrary member) also removes the set-order dependence.
- For a **triple point only**, drop when the three phases don't share one potential (spread `> _COEXIST_TOL`). Two-phase boundaries are exempt: a first-order boundary — where the stable phase jumps rather than crosses — carries a finite own-phase spread and is a real boundary, so clamping it to coexistence would wrongly drop it.

## Tests

`tests/unit/test_refine.py` gains direct `_dominated` cases: a non-coexisting own phase (dropped), an absent (`phi = +inf`) vertex (dropped), and the existing outside-dominator / equal-rival / triple cases, all order-independent.

---
_Generated by [Claude Code](https://claude.ai/code/session_011i7RNpdn7F3Kf6BvTfne8j)_

### DIFF
--- a/landau/refine.py
+++ b/landau/refine.py
@@ -190,39 +190,47 @@ def _state_row(phase: Phase, T: float, mu: float) -> dict:
     }
 
 
-# eV; largest spread in the own-phase potentials still read as one coexistence.
-# Comfortably above a converged triple/pair residual (~1e-5) and far below any
-# real domination gap, so it separates a genuine N-phase point from one where a
-# vertex doesn't belong.
+# eV; largest spread in a *triple*'s own-phase potentials still read as one
+# coexistence. Comfortably above a converged triple residual (~1e-5) and far
+# below any real domination gap. Not applied to two-phase boundaries, which may
+# be genuinely first-order (see _dominated).
 _COEXIST_TOL = 1e-3
 
 
 def _dominated(pt, phases: Mapping[str, Phase]) -> bool:
-    """True iff the refined transition ``pt`` is not a genuine coexistence and
-    should be dropped.
+    """True iff the refined transition ``pt`` is not a valid piece of the global
+    phase boundary and should be dropped.
 
-    Two failure modes:
+    A point is valid when its own phases are exactly the most stable set at
+    ``(pt.T, pt.mu)``. It is dropped when:
 
-    * Its own phases don't actually share a finite semigrand potential at
-      ``(pt.T, pt.mu)`` — one is absent (``phi = +inf``) or lies below the
-      others and merely rode along as a vertex of a Delaunay simplex, so the
-      claimed N-phase point isn't real. (This is how a pinned ordered phase
+    * An own phase is **absent** (``phi = +inf``): a phase that doesn't exist
+      here can't be a coexisting vertex. (This is how a pinned ordered phase
       reported absent outside its dome leaked into a triple point at ``c = 0``.)
-    * Some phase *outside* ``pt.phase_names()`` lies strictly below the
-      coexistence potential and is the true equilibrium.
+    * Some phase **outside** ``pt.phase_names()`` sits below the least-stable own
+      phase — then the own set isn't the top-|own| most stable, so this isn't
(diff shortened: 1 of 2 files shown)


landau/refine.py
```diff
@@ -191,16 +190,47 @@ def _state_row(phase: Phase, T: float, mu: float) -> dict:
     }
 
 
+# eV; largest spread in a *triple*'s own-phase potentials still read as one
+# coexistence. Comfortably above a converged triple residual (~1e-5) and far
+# below any real domination gap. Not applied to two-phase boundaries, which may
+# be genuinely first-order (see _dominated).
+_COEXIST_TOL = 1e-3
```
pmrv, line 197:
```suggestion
_TRIPLE_COEXIST_TOL = 1e-4
```

landau/refine.py
```diff
@@ -191,16 +190,47 @@ def _state_row(phase: Phase, T: float, mu: float) -> dict:
...
-    """True iff some phase outside ``pt.phase_names()`` has a lower
-    semigrand potential at ``(pt.T, pt.mu)`` — meaning the refined
-    transition we found isn't actually globally stable, so we drop it.
+    """True iff the refined transition ``pt`` is not a valid piece of the global
+    phase boundary and should be dropped.
+
+    A point is valid when its own phases are exactly the most stable set at
+    ``(pt.T, pt.mu)``. It is dropped when:
+
+    * An own phase is **absent** (``phi = +inf``): a phase that doesn't exist
```
pmrv, line 207: remove 'absent' comments.  CEF phases are not merged in yet, and might not even have this behavior anymore.

tests/unit/test_refine.py
```diff
@@ -1069,14 +1020,36 @@ def test_dominated_rival_equal_phi_returns_false():
...
     )
     pt = RefinedPoint(T=300.0, mu=0.0, phases=("A", "B", "D"))
-    assert _dominated(pt, phases) is False
+    assert _dominated(pt, phases) is True
+
+
+def test_dominated_absent_own_phase_returns_true():
+    """The Au-Cu degeneracy in miniature: a pinned ordered phase reported absent
+    (``phi = +inf``) must not survive as a triple-point vertex. C is absent, so
+    A+B+C is really just A+B and the triple point is dropped."""
```
pmrv, line 1056: more generic docstring.  In two weeks no one knows what Au-Cu refered to anymore

pmrv (commented):
@claude revise

landau/refine.py
pmrv: Applied in `d6db7c0`. Checked the headroom before tightening: the own-phase spread at the converged triple point of the eutectic fixture in `tests/unit/plot/test_triplepoint.py` is 2.05e-8 eV, so 1e-4 sits four orders above it. Comment above the constant now cites that number instead of the previous guess.

---

pmrv:
expand tests and rewrite docs according to your check

Issue #332, opened 2026-07-06
Phases take their free-energy parametrization indirectly (via interpolators + sampled line phases)


pmrv:
## The problem

Defining a phase's free energy is indirect. Instead of handing a `Phase` the function that *is* its free energy, you assemble it out of an interpolator strategy plus a bag of sample objects, and the phase reconstructs the free-energy curve internally.

Two concrete cases:

**`InterpolatingPhase`** takes a list of `AbstractLinePhase` terminals plus `num_coeffs` / `add_entropy` / `num_samples`, then decides the parametrization for you — `RedlichKister` if the endpoints are pure elements, `PolyFit` otherwise (`landau/phases/__init__.py:442-445`), refitting the terminals' `line_free_energy(T)` on every temperature (`_get_interpolation`). The caller never says "my free energy is this Redlich-Kister polynomial"; they hand over samples and hope the phase re-derives it.

**`Surface2DInterpolatingPhase`** is worse: to express a CALPHAD Gibbs surface you already have in closed form, you sample it onto a `(c, T)` grid as a list of `TemperatureDependentLinePhase` objects and hand those to a `surface_interpolator` (`CalphadSurface2DInterpolator`) that fits them *back* into a surface. `benchmarks/reproduce_sundman_aucu.py::_calphad_solution` is the whole dance — build `f(c, T)` analytically, sample it into line phases, then refit. A round-trip through sampling to get back something you started with.

The interpolator/parametrization is a mandatory constructor argument threaded indirectly, rather than an optional *builder* of the free-energy function. This makes phase construction verbose, couples every phase to the interpolator framework, and forces the sample → refit round-trip even when the free energy is known analytically. It also blocks the natural "just give me a callable" path that a `.tdb` importer (#138) or a bespoke model wants.

## Scope

Design/refactor, no rush — parking this as a design theme alongside the phase-subpackage cleanup (#137). A design stab follows in the first comment.

pmrv:
## Design stab

Make a solution phase take **the callable that is its free energy** and derive everything else from it; move the interpolator framework to the role of *building* that callable via classmethod constructors.

### Core

```python
@dataclass(frozen=True)
class FreeEnergyPhase(Phase):
    name: str
    free_energy: Callable        # (T, c) -> eV/atom, broadcasting
    dfdc: Callable | None = None # optional analytic dc-derivative; else numerical
```

[...]

PR #346, opened 2026-07-09
refactor(phases): compose CompoundEnergyPhase from Endmember/ExcessTerm objects

pmrv:
Stacked on #326. Assembles a `CompoundEnergyPhase` from small composable energy objects, in the spirit of the point-defect classes, instead of the flat single-excess-callable field.

## Structure (`landau/phases/compoundenergy.py`)

- **Sublattices are bare scalars** — `site_multiplicities`, a per-sublattice `a_s` summing to 1. No wrapper class.
- **`Endmember(occupation, energy)`** — a corner configuration (length-`S` 0/1 tuple) and its `°G_c(T)` callable (eV/atom).
- **`ExcessTerm`** (ABC) — each subclass implements `energy(y, T)`; `gradient(y, T)` defaults to a batched central finite difference and is overridden with the closed form where one exists. Every term names the sublattices it couples via a `sublattices` index tuple. Siblings:
  - `RegularSolutionExcess(sublattices, coefficient)` — `L(T)·Σ_{s∈sublattices} y_s(1-y_s)`, analytic gradient on those sublattices (zero elsewhere).
  - `CallableExcess(sublattices, function)` — an arbitrary `EG(y, T)` receiving the whole `y`; `sublattices` is its declared footprint; finite-difference gradient. The escape hatch for interactions without a bundled closed form (e.g. a CEF reciprocal term).

End-members and excess are **not per-sublattice**: an end-member is a corner (a product over all `S` sublattices) and an excess term couples whichever sublattices it names. The building blocks live in `compoundenergy.py:__all__` and are imported from the module, not re-exported to the top level (point-defect precedent).

## Structural validation (`__post_init__`)

Makes how/where excess terms attach explicit and forbids bad phases at construction:

- **End-members must cover every one of the `2**S` corners exactly once** — no missing or duplicated corner.
- **Excess terms must be compatible and unique** — every index in `range(S)` (no term referencing a sublattice the phase lacks), and no two terms may share a `(type, sorted sublattices)` key (no double-counted contribution). Same-type terms on *different* sublattices are distinct and allowed.

## Migration

`benchmarks/reproduce_sundman_aucu.py` and the `test_phases.py` CEF constructions build through the flat multiplicities + attached excess terms: the Au-Cu regular term (eq 8, over all four sublattices) is a `RegularSolutionExcess`, the reciprocal term (eq 9) a `CallableExcess`. The Assessment I diagram is unchanged — identical stable-point counts (`fcc 758, liquid 142, AuCu 90, AuCu3 77, Au3Cu 26`) against the #326 tip.

## Deferred

Reuse of the library interpolation classes' analytic gradients is deferred (per review): the note on the T/y/c derivative tension — `Interpolation.deriv()` serves whichever variable the fit is built in (T for entropy/enthalpy, y for the SCF stationarity, c for the reduced `f(c)`) — is recorded on the PR against #332.

## Tests

New `test_phases.py` cases: analytic `RegularSolutionExcess.gradient` vs finite difference; the base FD gradient as a drop-in; excess-term summation; analytic-vs-`CallableExcess` agreement in the SCF solve; and the structural rejections (missing/duplicate corner, out-of-range sublattice, duplicate `(type, sublattices)` key) plus same-type-different-sublattice acceptance.


landau/phases/compoundenergy.py
```diff
@@ -43,18 +52,124 @@
...
+    finite difference in every sublattice but is overridden with the closed form
+    where one exists.  ``y`` has its site fractions on the last axis and must
+    broadcast over any leading axes.
+    """
+
+    @abstractmethod
+    def energy(self, y, T):
+        """Excess energy for ``y`` of shape ``(..., S)`` -> ``(...)`` (eV/atom)."""
+
+    def gradient(self, y, T):
```
pmrv, line 130: Reeuse the interpolation classes and their gradients where we have them

pmrv (commented):
revise, make the structure and interaction clearer in docstrings and have a comment here too.  Excess terms and end members always come in pairs right?  the subkattice should group them then rather than just providing the multiplicity

landau/phases/compoundenergy.py
pmrv: Skipping the interpolation-gradient reuse for now, per your call. Noting the tension for later.

Three distinct derivatives are in play, and `Interpolation.deriv()` only serves the one whose variable the interpolation is built in:

- **∂/∂T** — entropy/enthalpy (`S = -∂G/∂T`). This is the derivative of the endmember `°G_c(T)` and the excess `L(T)`; those are 1-D fits *in T*, so their `.deriv()` gives it directly.
- **∂/∂y_s** — the site-fraction gradient the SCF stationarity uses (what this finite difference computes). An interpolation only helps here if the excess term is itself a fit *in y* (Redlich-Kister in `y_s`), not in T.
- **∂/∂c** — slope of the reduced `f(c)` curve (common tangent / chemical potential), where `c = Σ a_s y_s` is a linear combination of the `y`s.

So reusing the library gradients means matching each interpolation's variable to the derivative you need — a T-fit's `.deriv()` can't stand in for the y-gradient. Relevant to #332.

---

pmrv:
On what level the excess terms live — they don't sit at one uniform level, which is the crux of why they (and the end-members) are phase-level rather than under a `Sublattice`:

- The **regular** term (eq 8) is `L(T)·Σ_s y_s(1-y_s)` — summed over *all* sublattices; effectively phase-level.
- The **reciprocal** term (eq 9) couples *pairs* of sublattices `(r,s)` with the other two as spectators — sublattice-pair level.
- The general CALPHAD interaction parameter attaches to *one* mixing sublattice (a pair of interacting constituents on it) with the other sublattices fixed at specific constituents — so even the "per-sublattice" ones carry a dependence on the spectator sublattices' occupation.

And the end-members are the `2**S` corners, each a product spanning all sublattices. So neither end-members nor excess nest cleanly under a single `Sublattice`.

The faithful way to have the sublattice "group" things is the CALPHAD constituent-per-sublattice model: `Sublattice` carries its constituents (binary: A/B) + multiplicity, and the phase *derives* the corner end-members and the interaction parameters from them. That's the larger refactor. Otherwise I keep end-members/excess phase-level and just make the corner-product structure and the solver interaction explicit in the docstrings + an inline comment. Which do you want?

---

pmrv:
Done in d98c148:

- Dropped `Sublattice`; the phase takes `site_multiplicities` (plain `a_s` tuple) again.
- Each `ExcessTerm` now names its `sublattices` (the indices it couples), so attachment to the full space is explicit. `RegularSolutionExcess` acts on those (analytic gradient there, zero elsewhere); `CallableExcess` declares its footprint while receiving the whole `y`.
- `__post_init__` forbids bad phases structurally: end-members must cover every `2**S` corner exactly once (no missing/duplicate corner), and no excess term may reference a sublattice outside `range(S)` (incompatible) or share another's `(type, sublattices)` key (duplicate). Same-type terms on different sublattices are allowed.

Diagram unchanged (identical stable-point counts). Interpolation-gradient reuse left for later with the T/y/c tension note above.

---

PR #395, opened 2026-08-09
test(poly): correlate c/T/mu in poly_dataframe Hypothesis strategy

### LINKED ISSUE #70: Hypothesis strategies for polygon tests are terrible
Apparently [these guys](https://pygeoif.readthedocs.io/en/latest/) have some usable ones.

pmrv:
Closes #70.

`poly_dataframe` (`tests/unit/test_poly.py`) drew `c`, `T`, `mu`, `phase`, and `phase_id` independently per row. As noted in the #116 backlog comment, this means generated rows have no relation between concentration and thermodynamic state, so the "phase diagram"-ness of the fuzzed input is fictional — fine for crash-testing, but it limits what regressions the property tests can catch.

## Change

Each `(phase, phase_unit)` group is now built by `_phase_region`, which traces a smooth curve in a scalar parameter `t`:

- `T = T0 + T_amp * t`
- `mu = mu0 + mu_amp * (2*t - 1)`
- `c = c0 + c_amp * sin(2*pi*t)`

so `c`, `T`, and `mu` correlate with each other instead of being independent, and border points trace a coherent region outline (as `calc_phase_diagram` output would) rather than scattered noise. `phase_id` is now derived from `phase`/`unit` instead of sampled independently of `phase`. Each phase gets one or two disconnected regions (simulating a phase stable in separate windows).

`Segments` (and its TSP variants) reduce border points via `get_transitions`, which pairs rows sharing `(mu, T)` across phases into a coexistence point — a few such pairs are still forced between the first region of each phase, generalizing the old single hardcoded pair to real generated boundary values.

I looked at pygeoif's Hypothesis strategies (the issue's suggestion) but they generate arbitrary valid geometries and don't encode the domain-specific correlation this test actually needs (`c` tied to `mu`/`T`, and `Segments` needing genuine shared-`(mu, T)` coexistence rows), so I wrote a domain-specific composite strategy instead and kept the test suite dependency-free.

## Testing

- `pytest tests/unit/test_poly.py -q` — 42 passed, run 3x with `--hypothesis-seed=random` for stability
- `pytest -q` (full suite) — 711 passed, 1 xfailed (pre-existing `Segments × μ-T` xfail)
- `ruff check tests/unit/test_poly.py` — no new findings (5 pre-existing import-order/unused-import findings on `main`, unrelated to this change, left untouched)

---
_Generated by [Claude Code](https://claude.ai/code/session_01Rp6TcW6XxZrg74BYJu8RKd)_

### DIFF
--- a/tests/unit/test_poly.py
+++ b/tests/unit/test_poly.py
@@ -38,66 +38,84 @@ except ImportError:
 
 
 @st.composite
+def _phase_region(draw, phase, unit):
+    # One (phase, phase_unit) group, traced as a smooth curve in a scalar
+    # parameter `t` so that c, T and mu correlate with each other instead of
+    # being drawn independently -- border points then look like a plausible
+    # phase-region outline (as calc_phase_diagram would produce) rather than
+    # unstructured noise.
+    n_border = draw(st.integers(min_value=4, max_value=10))
+    n_interior = draw(st.integers(min_value=0, max_value=3))
+    n_points = n_border + n_interior
+    t = np.sort(
+        draw(
+            st.lists(
+                st.floats(min_value=0, max_value=1, allow_nan=False, allow_infinity=False),
+                min_size=n_points,
+                max_size=n_points,
+                unique=True,
+            )
+        )
+    )
+
+    T0 = draw(st.floats(min_value=0, max_value=800, allow_nan=False, allow_infinity=False))
+    T_amp = draw(st.floats(min_value=0, max_value=200, allow_nan=False, allow_infinity=False))
+    mu0 = draw(st.floats(min_value=-8, max_value=8, allow_nan=False, allow_infinity=False))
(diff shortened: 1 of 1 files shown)


pmrv:
`c = c0 + c_amp * sin(2*pi*t)`  sounds like we could open ourselves to unphysical edge cases.  Use a ideal solution type sigmoid.

pmrv:
Switched to a sigmoid: `c = c_lo + (c_hi - c_lo) / (1 + exp(-steepness * (t - 0.5)))`, matching `IdealSolution.concentration`'s `1 / (1 + exp(...))` form. It's monotonic and bounded in `(c_lo, c_hi) ⊂ (0, 1)` by construction, so it can't double back the way the sinusoid could. `T` and `mu` are unchanged (already monotonic affine functions of `t`).

Pushed in 29ff5bb. Full suite still green (711 passed, 1 pre-existing xfail).

---

